PR #114, opened 2015-02-09
Mark failing test as skipped

ss23:
This is better than tests that fail.


tractorcow:
Yeah it wasn't really a good test.

Issue #92, opened 2014-11-11
Add locking to deployments such that two people can't deploy at once


ss23:
Not high priority because developer workflows generally avoid this, but it would be wonderful to have it built in.

mandrew:
+1 for this as it happened to me on friday which was frustrating : (

halkyon:
This has been merged into master.

PR #116, opened 2015-03-01
BUG Disallow simultaneous deployments (#92)

halkyon:
Works the same way as `DataTransferJob`, which currently stops anyone from
being able to perform transfers at once.


ss23:
I'm wary that old jobs in a non finished state will break this as soon as we deploy. Can you implement a mechanism for bypassing this if you need to *or* jobs that are older than a certain amount of time (30 minutes) not counting for this?

halkyon:
@ss23 Good idea. I've added that to the filters.

stojg:
There are some errors, I've created a new PR at https://github.com/silverstripe/deploynaut/pull/117

PR #122, opened 2015-04-01
Remove Individual member assigning for environments.

brettt89:
Remove Individual member assigning for environments. 
Use Group / Parent Group for organising Roles.

Use Roles attached to Groups for Permissions.

E.g.
{Parent Group} -> {Group} -> {Role}
MPIGOVT -> Instance Manager -> IM_ACCESS
MPIGOVT -> Backup Instance Manager -> BIM_ACCESS

Assign Group to Environment Permissions.
E.g.
CanDeploy -> MPIGOVT-Instance Manager, MPIGOVT-Backup Instance Manager

When addin a new user, only need to add them to the corresponding group within the parent.
(does not need to be added to parent).
E.g.
{Member} -> {Group}
John Doe -> MPIGOVT-Instance Manager

### DIFF
--- a/code/model/DNEnvironment.php
+++ b/code/model/DNEnvironment.php
@@ -97,24 +97,15 @@ class DNEnvironment extends DataObject {
 	 * @var array
 	 */
 	public static $many_many = array(
-		"Viewers"            => "Member", // Who can view this environment
-		"ViewerGroups"       => "Group",
-		"Deployers"          => "Member", // Who can deploy to this environment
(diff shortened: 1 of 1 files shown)


ss23:
I am unsure whether this is a good change.

For smaller users of deploynaut, being able to assign single users is a good thing. This is also used internally at SilverStripe Bespoke.

Changing the UI, making it hidden if you choose to hide it, I'm okay with, but otherwise I think this is a bad change.

Also, tests are failing!

PR #123, opened 2015-04-07
ENHANCEMENT Allow menu to be extendable from DNProject

halkyon:
Allows an extension class to add their own menu items and to remove
existing ones if necessary.

I'm not happy about the `Controller::curr()` usage, so if there's a better
way of doing this, let me know :)


halkyon:
I've forgotten a bunch of other templates that needed updating. I'll close and re-open later.

PR #124, opened 2015-04-07
ENHANCEMENT Allow menu to be extendable from DNProject

halkyon:
Allows an extension class to add their own menu items and to remove
existing ones if necessary.

### DIFF
--- a/code/model/DNProject.php
+++ b/code/model/DNProject.php
@@ -148,6 +148,33 @@ class DNProject extends DataObject {
 	}
 
 	/**
+	 * Get the menu to be shown on projects
+	 *
+	 * @return ArrayList
+	 */
+	public function Menu() {
+		$list = new ArrayList();
+
+		$list->push(new ArrayData(array(
+			'Link' => sprintf('naut/project/%s', $this->Name),
+			'Title' => 'Deploy',
+			'IsActive' => Controller::curr()->getAction() == 'project'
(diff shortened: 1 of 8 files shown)


code/model/DNProject.php
```diff
@@ -148,6 +148,33 @@ public function HasDiskQuota() {
...
+	public function Menu() {
+		$list = new ArrayList();
+
+		$list->push(new ArrayData(array(
+			'Link' => sprintf('naut/project/%s', $this->Name),
+			'Title' => 'Deploy',
+			'IsActive' => Controller::curr()->getAction() == 'project'
+		)));
+
+		if(Controller::curr()->FlagSnapshotsEnabled()) {
```
tractorcow: What if the current controller isn't DNRoot (e.g. in the case of a nested controller?). How about `singleton('DNRoot')` or just making it a static method?

halkyon: Static seems fine, and it's not used anywhere else from what I can tell :) have rebased the change.

stojg:
I think it looks good enough to support adding more menu items, which we will have to do unless we want to bundle everything into deploynaut base module, which I don't like at all.

In the end I think we will have to do some UI refactoring around the navigation and presentation. Deploynauts look and feel could need some :heart:

tractorcow:
Nice job. :)

PR #129, opened 2015-04-16
Hardening

mateusz:


code/backends/CapistranoDeploymentBackend.php
```diff
@@ -64,12 +64,53 @@ public function deploy(DNEnvironment $environment, $sha, DeploynautLogFile $log,
...
+		$cleanupFn = function() use ($self, $projectName, $environmentName, $args, $env, $log) {
+			$command = $self->getCommand('deploy:cleanup', 'web', $projectName.':'.$environmentName, $args, $env, $log);
+			$command->run(function ($type, $buffer) use($log) {
+				$log->write($buffer);
+			});
+			if(!$command->isSuccessful()) {
+
+				// Notify ops via email, if possible.
+				$from = Config::inst()->get('EmailMessagingService', 'default_from');
+				$to = Config::inst()->get('EmailMessagingService', 'ops_to');
```
halkyon: I would probably call this something else, not `ops_to`, as that's very specific. Perhaps it should also be configured from the ss_environment file?

PR #132, opened 2015-05-26
Fixing performance issues when viewing deploynaut data archives

halkyon:
Each `DNDataArchive::canView()` call checks whether the archive can be
either downloaded or restored. This check is done against the
DNEnvironment associated with that archive, but it's only needed
to be called once, not for each archive.

### DIFF
--- a/code/model/DNDataArchive.php
+++ b/code/model/DNDataArchive.php
@@ -79,6 +79,10 @@ class DNDataArchive extends DataObject {
 		),
 	);
 
+	private static $_cache_can_restore = array();
+
+	private static $_cache_can_download = array();
+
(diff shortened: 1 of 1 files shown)


ss23:
:+1:

halkyon:
@ss23 :shipit: ? :D

ss23:
> Failed — The Travis CI build failed

halkyon:
@ss23 Ah, of course, if `$member` is an object that can't be used as an array key ;-)
Fixed and squashed, let's check the next build passes!

ss23:
:bug: :hammer:

halkyon:
All is well!

PR #138, opened 2015-06-10
Rename sync task so it doesn't clash with the CMS.

mateusz:


ss23:
What's up with that weird Travis failure?

mateusz:
I've pushed the branch by mistake to the root repo, before realising Im not pushing to a fork. That branch is now removed - it looks like github oddity triggered by branches that match by name. I'd say safe to ignore.

PR #145, opened 2015-06-22
Fixes for catching errors from attachFile and logging them.

halkyon:
This currently doesn't log the exact error, if there was one.

### DIFF
--- a/code/model/DNDataArchive.php
+++ b/code/model/DNDataArchive.php
@@ -317,10 +317,9 @@ class DNDataArchive extends DataObject {
 	 *
 	 * @param string $sspakFilepath
 	 * @param DNDataTransfer $dataTransfer
-	 * @param DeploynautLogFile $log
 	 * @return bool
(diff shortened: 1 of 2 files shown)


code/model/DNDataArchive.php
```diff
@@ -317,10 +317,9 @@ public function generateFilepath(DNDataTransfer $dataTransfer) {
 	 *
 	 * @param string $sspakFilepath
 	 * @param DNDataTransfer $dataTransfer
-	 * @param DeploynautLogFile $log
 	 * @return bool
 	 */
-	public function attachFile($sspakFilepath, DNDataTransfer $dataTransfer, DeploynautLogFile $log) {
+	public function attachFile($sspakFilepath, DNDataTransfer $dataTransfer) {
```
stojg: This method is still calling $log ?

halkyon: nope

PR #146, opened 2015-06-22
Small cleanups to Process

halkyon:
- Use the second argument to Process instead of `cd` in the command
 - Concat strings in a nicer way using sprintf

### DIFF
--- a/code/backends/CapistranoDeploymentBackend.php
+++ b/code/backends/CapistranoDeploymentBackend.php
@@ -280,9 +280,9 @@ class CapistranoDeploymentBackend extends Object implements DeploymentBackend {
 			$log->write(sprintf('Backup of assets from "%s" done', $name));
 		}
 
-		$log->write('Creating *.sspak file');
+		$log->write('Creating sspak file');
 		$sspakFilename = sprintf('%s.sspak', $dataArchive->generateFilename($dataTransfer));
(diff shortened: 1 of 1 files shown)


code/backends/CapistranoDeploymentBackend.php
```diff
@@ -340,19 +340,18 @@ public function rebuild(DNEnvironment $environment, $log) {
...
 	protected function dataTransferRestore($workingDir, DNDataTransfer $dataTransfer, DeploynautLogFile $log) {
 		$environment = $dataTransfer->Environment();
 		$name = $environment->getFullName();
 
 		// Rollback cleanup.
-		$self = $this;
-		$cleanupFn = function() use($self, $workingDir, $environment, $log) {
+		$cleanupFn = function() use($workingDir, $environment, $log) {
 			// Rebuild makes sense even if failed - maybe we can at least partly recover.
 			$self->rebuild($environment, $log);
```
stojg: you removed $self from the anon func use statement, does this really work?

halkyon: yes, `$self` was copy/pasta when someone else originally did this.

halkyon: fixed :)

PR #147, opened 2015-06-22
Move permission fixing to when a user uploads their snapshot

halkyon:
Currently the permissions are fixed on restore action, but this makes
more sense to fix the problem at the source, which is when a user
uploads their snapshot.


micmania1:
All good but I don't have permission to merge :(

stojg:
Still a bit annoying that the commands are set with the argument instead of using a `ProcessBuilder` that escaped all the things.

But it's still a good change doing this on receiving a sspak instead of before every time a sspak is 'restored'.

PR #149, opened 2015-06-30
Sanitise the folder names with DNDataArchive consistently

halkyon:
Framework uses FileNameFilter for sanitising the folder names when
generating them from a value with special characters like spaces, etc,
but the current rules don't work as they replace spaces with underscores
and such. We should rely on FileNameFilter::filter() to do the
sanitising work for us instead of hardcoding those regex rules in
DNDataArchive.

### DIFF
--- a/code/model/DNDataArchive.php
+++ b/code/model/DNDataArchive.php
@@ -288,19 +288,12 @@ class DNDataArchive extends DataObject {
 	 */
 	public function generateFilename(DNDataTransfer $dataTransfer) {
 		$generator = new RandomGenerator();
-		$sanitizeRegex = array('/\s+/', '/[^a-zA-Z0-9-_\.]/');
-		$sanitizeReplace = array('/_/', '');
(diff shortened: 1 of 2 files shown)


ss23:
Is it the case you can create two different projects with overlapping names?

halkyon:
@ss23 Yep. It currently creates 2 folders, because some later code uses `Folder::find_or_make()` which uses `FileNameFilter`. With this change it forces the generated path to be pre-filtered using the same filter rules used in the rest of the framework for files and folders.

stojg:
Come on @ss23, I know you want to merge this.. ;)

ss23:
Hey, it's a good change, I just didn't want to jump ahead of anyone else, as it looked like it might have some unintended consequences that I haven't thought through.

ss23:
@halkyon Will this break existing snapshots, or is that filter 'basically' the same? I saw a test changed? Do we need to rename folders to make it work?

halkyon:
Thanks guys :)

halkyon:
@ss23 They'll get a new path, but it shouldn't break existing snapshots as they'll use the old path. I'll double check though.

PR #150, opened 2015-06-30
Add better logging for when git clone fails

halkyon:
Currently if there's a problem running `git clone` it throws an error
but doesn't log this anywhere. This change sends the error output to
`DEPLOYNAUT_LOG_PATH/clonegitrepo.log`


stojg:
:shipit:

PR #152, opened 2015-07-01
Use `tar` to validate what files are in the DataArchive

stojg:
This speeds up the validation process immensely when dealing with multiple GB snapshots and doesn't use any more diskspace


halkyon:
@stojg Tests fail because you used short-hand array syntax. Probably should fix that until we ditch 5.3 support in framework.

tests/DNDataArchiveTest.php
```diff
@@ -117,4 +117,58 @@ public function testCanMoveTo() {
...
+		$result = $archive->validateArchiveContents('db');
+		$this->assertTrue($result->valid());
+	}
+
+	public function testValidateArchiveContentsDBFails() {
+		$archive = DNDataArchive::create();
+		$archive->ArchiveFile()->Filename = __DIR__.'/sspaks/assets.sspak';
+		$result = $archive->validateArchiveContents('db');
+		$this->assertFalse($result->valid());
+		$this->assertEquals('The snapshot is missing the database.', $result->messageList()[0]);
```
stojg: It's actually these guys failing the tests

stojg:
Finally the test pass through, I had to bump them until various network failures stopped messing with composer.

PR #153, opened 2015-07-01
Fixing a regression from 29f1929835 where file hash isn't created.

halkyon:


halkyon:
This raises another question though: hashing a large snapshot can take a while and be an expensive operation. Is it worth doing such a thing?

stojg:
Are we using the ArchiveFileHash for anything?

halkyon:
@stojg I don't know, couldn't find any references. Anyone else know?

stojg:
it’s an regression, so from not working to working as bad as before.. is better.. I’ll merge

PR #157, opened 2015-07-02
Fixing performance. Should check environment ID not archive ID.

halkyon:
Partially reverts bd458022, doesn't actually need to be checking `$this->ID`
as it doesn't use it as part of the permission check.

Additionally, it didn't actually check the logged in Member ID
properly, so the cache key broke.


halkyon:
@micmania1 I reverted your commit, sorry. This broke the snapshot listing performance. Those `can()` methods don't need to be checked for every archive, it just needs to check once for the member, hence why the cache key is only for the current member ID and doesn't include the archive.

Did you find a specific case where this breaks the snapshots listing somewhere?

PR #158, opened 2015-07-14
NEW Added extension hook to getProcessEnv

micmania1:
Added an extension hook so you can replace the GIT_SSH file with a custom one and add custom key/values.


stojg:
after tests passes :shipit:

PR #159, opened 2015-07-16
Added methods to easily obtain keys

micmania1:


halkyon:
Could there be cases where current projects have keys directories with underscores that `FileNameFilter` will try looking for dashes on instead?

code/model/DNProject.php
```diff
@@ -509,10 +509,62 @@ public function onAfterDelete() {
 	 * Fetch the public key for this project.
 	 */
 	public function getPublicKey() {
+		$filter = FileNameFilter::create();
```
halkyon: Couldn't this method just be:

```
$path = $this->getPublicKeyPath();
return $path ? file_get_contents($path) : null;
```

micmania1: I don't think so. This filters any harmful names (ie ../other-project/ would be other-project).

PR #160, opened 2015-07-20
FIX Update DNProject to use better detection for private key.

micmania1:


halkyon:
The travis error seems unrelated "zlib_decode()". Merging.

PR #162, opened 2015-07-20
BUG: Old deployments show an "Invalid" status when viewing the log

stojg:


stojg:
Please hold mergin this PR for a while

PR #164, opened 2015-07-22
Remove cleanup alerts_to, it doesn't get used. Replace with hooks.

halkyon:
This sort of functionality is better suited in a module, not in the core deploynaut.
`alerts_to` isn't being used on any of our implementations of this module, probably because it was too noisy :)


ss23:
:+1:

stojg:
:trumpet:

PR #169, opened 2015-07-25
tailing spaces etc.

spekulatius:
just codestyle like tailing spaces etc.


ss23:
:+1: 
We should probably make scrutinizer pick up on these and prevent them getting in the codebase in the first place

spekulatius:
yupp, somehow it would be nice to keep it clean from simple code styling issues like tailing spaces etc.

PR #184, opened 2015-07-29
Add Repository Interface detection

hafriedlander:


code/model/DNProject.php
```diff
@@ -621,6 +621,78 @@ public function getRepositoryURL() {
...
+	 */
+	public function getRepositoryInterfaceName() {
+		$url = parse_url($this->CVSPath);
+		$interfaces = $this->config()->repository_interfaces;
+		$host = strtolower($url['host']);
+
+		if (array_key_exists($host, $interfaces)) {
+			$interface = $interfaces[$host];
+
+			return isset($interface['name']) ? $interface['name'] : ucfirst(preg_split('/\./', $host)[0]);
```
micmania1: `preg_split('/\./', $host)[0]` I don't think you can do this in 5.3

code/model/DNProject.php
```diff
@@ -621,6 +621,78 @@ public function getRepositoryURL() {
...
+	 * {configuration} can either be boolean true to auto-detect both the host and the
+	 * name of the UI provider, or a nested array that overrides either one or both
+	 * of the auto-detected valyes
+	 *
+	 * @var array
+	 */
+	static private $repository_interfaces = array(
+		'github.com' => true,
+		'bitbucket.org' => true,
+		'repo.or.cz' => array(
```
micmania1: is this supposed to be here?

code/model/DNProject.php
hafriedlander: Ah, true. Will fix.

code/model/DNProject.php
hafriedlander: repo.or.cz is the third most common free git hosting service. Happy to take it out, but ideally this'd support all the common public hosts by default.

micmania1: fair enough. I've obviously been living under a rock :)

hafriedlander: The opposite - it's _super_ old (long before github) :).

PR #186, opened 2015-07-29
Plaform forms1

clarkepaul:
Additional forms styling for the login and create snapshots and upload


micmania1:
Tidied up in #188

PR #189, opened 2015-07-29
Cleaned up snapshot log template

micmania1:


code/model/jobs/DNDataTransfer.php
```diff
@@ -89,7 +89,7 @@ public function Link() {
 	}
 
 	public function LogLink() {
-		return $this->Link() . '/log';
+		return $this->Link();
```
hafriedlander: Why keep the separate LogLink function if it's no different to Link now?

micmania1: I've updated things a bit so LogLink is useful again.

PR #190, opened 2015-07-30
Added overview tab and moved into an include

micmania1:


templates/Includes/DeploymentTabs.ss
```diff
@@ -0,0 +1,12 @@
+<ul class="nav nav-tabs">
+    <li<% if $Overview %> class="active"<% end_if %>><a href="$CurrentProject.Link">Overview</a></li>
```
hafriedlander: I'd probably have called this `$OverviewActive` or something slightly more specific, but not enough of an issue to block merge.

PR #195, opened 2015-07-30
ran npm install

micmania1:


ss23:
Shouldn't this be handled like composer?

hafriedlander:
@ss23 It is - package.json defines the dependancies. This is the complied css after updating the dependancies to the latest (basically, fontawesome 4.3 => 4.4)

ss23:
What I meant was, why would we package the results of a npm install? Shouldn't we *not* package it, like we don't package the results of a composer install?

hafriedlander:
We aren't. This is the _compiled_ outcome. npm install puts everything under the node_modules folder (which hasn't been committed).

hafriedlander:
(However it's probably going to need to be re-done once all the other PRs are merged. Checking in complied css == merge headaches)

micmania1:
Paul made the same change and its been merged.

PR #199, opened 2015-07-30
Dont fatal error when git repo not yet downloaded

hafriedlander:


code/model/DNReferenceList.php
```diff
@@ -59,7 +59,12 @@ public function setLimit($limit) {
 	 * @return array()
 	 */
 	protected function getReferences() {
-		$repository = new Gitonomy\Git\Repository($this->project->LocalCVSPath);
+		try {
+			$repository = new Gitonomy\Git\Repository($this->project->LocalCVSPath);
+		} catch (Exception $e) {
```
stojg: Is it possible to log the error somehow?

halkyon: @stojg If the repo couldn't be cloned or fetched, it should be logged anyway, so it doesn't matter?

stojg: fair enough, as long as it's logged somewhere that the repo is borked.

PR #209, opened 2015-07-31
Make support links configurable with config system

hafriedlander:


micmania1:
The CMS isn't included in deploynaut so this won't work. Can you change it to a global template provider? I'll close for now incase somebody merges.

PR #212, opened 2015-07-31
FIX Select2 was being called on dropdowns before they were visible so…

micmania1:
… width wasn't correctly set

### DIFF
--- a/javascript/deploynaut.js
+++ b/javascript/deploynaut.js
@@ -251,6 +251,14 @@
 			if(!confirm(msg)) e.preventDefault();
 		});
 
+		$('.upload-exceed-link').on('click', function() {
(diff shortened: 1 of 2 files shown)


javascript/deploynaut.js
```diff
@@ -251,6 +251,14 @@
 			if(!confirm(msg)) e.preventDefault();
 		});
 
+		$('.upload-exceed-link').on('click', function() {
+			$(this).tab('show');
+
+			var id = $(this).attr('data-target');
+			$(id).find('select').select2();
```
hafriedlander: This calls select2 every time the link is clicked - is this OK?

micmania1: yep, it just works like a refresh.

PR #216, opened 2015-07-31
Added style to postsnapshot template

micmania1:


sass/style.sass
```diff
@@ -13,6 +13,10 @@ $textarea-transition-time: 0.6s
 @import "_form.sass"
 @import "_type.sass"
 
+ul > li > ul
+	margin-bottom: 0
+	margin-top: 0
+
```
hafriedlander: This seems very general - how much did you check other pages for regressions?

clarkepaul: I think I need to look at the ul stuff in general as i made it have a negative margin on top so it is associated to the text above. I'll look into it separately. To me this looks good but the "margin-bottom" might still need some space.

micmania1: I think this might have affected the deploy form styles. I'll close until its resolved.

PR #223, opened 2015-07-31
allow_web_editing true by default

shoaibali:
I have given this some thought, either we allow_web_editing by default to be true or we refactor this code (https://github.com/silverstripe/deploynaut/blob/master/code/model/DNEnvironment.php#L902) so that, for the very first time when deploynaut is setup the administrator sees an option to  "create capistrano deploy configuration".

### DIFF
--- a/_config/dnroot.yml
+++ b/_config/dnroot.yml
@@ -31,3 +31,5 @@ Injector:
     properties: 
       BaseDir: "../deploynaut-resources/build-cache"
       CacheSize: 5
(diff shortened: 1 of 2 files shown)


spekulatius:
I would enable it by default as well. The backend is for admins only - especially in deploynaut :+1:

spekulatius:
@sminnee are you happy with this being split into different pull requests? Would you merge it then?

PR #228, opened 2015-08-02
breaking down some over long lines

spekulatius:


spekulatius:
I rebased this one, it should be good to go now @micmania1

PR #237, opened 2015-08-03
back link on the deploy screen

clarkepaul:
Link allows the user to go back to the environment screen (from the deploy screen) if they can deploy

### DIFF
--- a/templates/Layout/DNRoot_deploy.ss
+++ b/templates/Layout/DNRoot_deploy.ss
@@ -13,7 +13,17 @@
 <div class="content">
 	<div class="row">
 		<div class="col-md-12 environment-details">
-			<h4>Environment Details:</h4>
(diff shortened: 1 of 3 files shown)


templates/Layout/DNRoot_deploy.ss
```diff
@@ -13,7 +13,17 @@
 <div class="content">
 	<div class="row">
 		<div class="col-md-12 environment-details">
-			<h4>Environment Details:</h4>
+			<% with $CurrentProject %>
+				<% if $DNEnvironmentList %>
+					<% loop $DNEnvironmentList %>
+						<% if $IsSection %><% if CanDeploy %>
```
halkyon: CanDeploy isn't needed here, it's just a link back to the environment.

PR #239, opened 2015-08-03
Template and style changes to environment details area to include right links

clarkepaul:
Please visually test as I can't test locally


micmania1:
This was merged here: https://github.com/silverstripe/deploynaut/pull/251

PR #240, opened 2015-08-03
Fixing performance issue where Member::currentUser() is queried a lot.

halkyon:
This should be cached, so the obvious fix is to keep the first record in
an in-memory cache and refer to that for other calls to canView() in the
same request.

### DIFF
--- a/code/model/DNProject.php
+++ b/code/model/DNProject.php
@@ -76,6 +76,14 @@ class DNProject extends DataObject {
 	protected static $relation_cache = array();
 
 	/**
+	 * Member::currentUser() cache to avoid re-query the same thing
+	 * again and again.
+	 *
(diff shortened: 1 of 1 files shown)


micmania1:
What happens when you comment out the admin check?

code/model/DNProject.php
```diff
@@ -76,6 +76,14 @@ class DNProject extends DataObject {
 	protected static $relation_cache = array();
 
 	/**
+	 * Member::currentUser() cache to avoid re-query the same thing
+	 * again and again.
+	 *
+	 * @var mixed
+	 */
+	private static $_cache_member = null;
```
stojg: can you maybe rename it to $_cache_current_member? otherwise it sounds like a generic hashed member cache

PR #243, opened 2015-08-03
Don't show deploy form if you don't have permissions.

halkyon:
Links should still work, as you can view the environment, just not deploy to it.


halkyon:
@stojg Oops, fixed that weird commit there ;-)

PR #248, opened 2015-08-04
Memory cache the current logged in member when deciding which environment the member can view

stojg:


code/model/DNProject.php
```diff
@@ -358,9 +363,19 @@ public function getRepository() {
 	 * @return ArrayList
 	 */
 	public function DNEnvironmentList() {
+
+		if(!self::$_current_member_cache) {
+			self::$_current_member_cache = Member::currentUser();
+		}
+
+		if(self::$_current_member_cache === false) {
+			new ArrayList();
```
halkyon: not assigned to anything here?

stojg:
In my local environment with a lot of environment it decreases the # SQL from 585 to 258, ie removing ~330 duplicate queries

code/model/DNProject.php
stojg: sloppy, Ill fix

halkyon: The code here is a little confusing. Couldn't you combine the `!` and `false` check into one?

stojg:
In worst case if there is no logged in member it will still do all those queries, but a user shouldn't get this far without a login.

halkyon:
ok nevermind my last comment about confusing. It makes sense now that you return an ArrayList :)

PR #250, opened 2015-08-04
Memory cache the result of DNRoot::getCurrentProject()

stojg:


stojg:
This will help the various calls from the templates and other callers

PR #265, opened 2015-08-05
Phpdoc 2

spekulatius:


stojg:
Thanks! would you mind rebase the three commits into one commit?

code/tasks/SyncProjectsAndEnvironments.php
```diff
@@ -2,6 +2,9 @@
 
 class SyncProjectsAndEnvironments extends BuildTask {
 
+	/**
+	 * @param  SS_HTTPResponse $request
```
halkyon: Double space between param and SS_HTTPResponse? :)

spekulatius: Sorry, I'll fix that - including the multiple commits. Give me a sec.

spekulatius:
Hello @stojg and @halkyon ,

Thanks for your reply. I fixed this - finally. Sorry for the multiple commits and the double spaces. Shouldn't be like this.

Cheers

PR #270, opened 2015-08-06
Check if content exists before calling file_get_contents()

halkyon:


code/control/DeploynautLogFile.php
```diff
@@ -14,17 +15,29 @@ public function getLogFile() {
...
 		return file_exists($this->logFile);
 	}
 
+	/**
+	 * Return the content of the log file.
+	 * @return string
+	 */
 	public function content() {
-		return file_get_contents($this->logFile);
+		return $this->exists() ? file_get_contents($this->logFile) : '';
```
stojg: Can we reply that this log doesn't exists? otherwise it might look like a bug and someone needs to log into the server and check if it's there or something else has crashed..

halkyon: Do you mean log that it doesn't exist?

halkyon: @stojg Added a message if it doesn't exist :)

PR #284, opened 2015-08-17
Fix issue where scroll bars on main window appear when log shows

dhensby:


sass/style.sass
```diff
@@ -587,7 +587,8 @@ button.sidebar-open
 	max-height: 450px
 	background: #333
 	color: #fff
-	margin-top: 0
+        margin-top: 0
```
halkyon: Tabs not spaces :)

dhensby: bloody vim

halkyon:
You'll also need to compile the CSS and commit the resulting file as well, since you're just changing an SCSS file here, which does nothing by itself :) `cd deploynaut && npm update && npm build`

dhensby:
hmm - literally just did that and no changes...

dhensby:
OK, so `npm install` goes and installs all the deps I need, but `npm update` and `npm build` do nothing :/

dhensby:
meh, just did `npm install -g node-sass && node-sass -r ./sass/style.sass -o ./css/` and it worked

dhensby:
Whoop - first contribution to 'naut. I'm practically a cloudbuster now!

ss23:
Dammit @dhensby I knew there was a reason our cloud server went down, YOU BUSTED IT.

dhensby:
Hot-diggity-dog! I knew that `margin-bottom` enabled a hack that allowed us to calculate the number of seconds to offset UTC by to ensure deploys started correctly and allowed us to prevent race conditions on deployments!!!

PR #294, opened 2015-08-19
Automatically scroll the log to the bottom

dhensby:
When the deploy log is updated, we should scroll the div to the bottom to see the latest updates.

### DIFF
--- a/javascript/deploynaut.js
+++ b/javascript/deploynaut.js
@@ -73,16 +73,36 @@
 	}
 
 	var queue = {
+		autoScroll: true,
 		showlog: function(status, content, logLink) {
 			var self = this;
+			//add scroll listener
+			content.on('scroll', function(ev) {
+				// if we are scrolled to the bottom then autoScroll should be on otherwise we've scrolled somewhere else
+				// and we shouldn't move the scroll any more
+				if (content.scrollTop() >= (content[0].scrollHeight - content.innerHeight())) {
+					self.autoScroll = true;
+				}
+				else {
+					self.autoScroll = false;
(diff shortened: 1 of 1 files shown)


stojg:
yes, I like this a lot! I'm going to see if I can get scrutinizer to behave properly by kicking it.

dhensby:
this does kind of need testing...

dhensby:
also, if the user manually scrolls we should probably turn it off, otherwise it'd get annoying

stojg:
Yeah, we need to capture a user scroll event and then autoscroll. I think that was what I've tried doing 2 years ago or so, but it fell between the chairs.

javascript/deploynaut.js
```diff
@@ -79,6 +79,8 @@
 			function(data) {
 				status.text(data.status);
 				content.text(data.content);
+                //scroll the content to the bottom
```
halkyon: We-ooh-wee-ohh. The indentation police are here to tell you that these spaces should be tabs ;-)

dhensby:
@stojg done - If you scroll and you're not at the bottom, then it won't scroll for you. if you scroll to the bottom, it'll keep going to the bottom.

@halkyon fixed :P stupid IDE

dhensby:
Also, this approach allows us to easily add a button to turn auto scrolling on/off :)

dhensby:
Only other thing I can think is that the scroll event could be namespaced in case some other scroll event is attached and we accidently detach that with the `off` call

dhensby:
Whoop

PR #296, opened 2015-08-25
Multi step deploy version 2

stojg:


code/control/DNRoot.php
```diff
@@ -885,6 +887,60 @@ public function getDeployForm() {
...
+             new HiddenField('SelectRelease', 'SelectRelease', 'SHA'),
+             new HiddenField('SHA', 'SHA', $selectedBuild)
+         ));
+
+		// clear out the old action button(s) from the 'choose revision' view
+		$form->setActions(new FieldList());
+
+		// ask the environments backend to amend the deploy form so that
+		// they can later be passed back via the resque job
+		$environment->Backend()->setDeployStrategies($form, $environment);
```
stojg: In this stage of the implementation it's expected that the backend updates the form like this

```
$form->setActions(new FieldList(
			FormAction::create('doDeploy', "Deploy")
				->addExtraClass('btn btn-primary deploy-button')
				->setAttribute('data-environment-name', Convert::raw2att($environment->Name))
		));
		$form->Fields()->push(new ReadonlyField('ReadonlySHA', 'SHA', $form->Fields()->fieldByName('SHA')->attrValue()));
```

but any other strategy would work

mateusz: I think @micmania1  should review this, because he is working on that bit, unless he is not. I think the $form was removed from the planDeploy?

javascript/deploy.js
```diff
@@ -0,0 +1,227 @@
...
+
+	/**
+	 * this is a jQuery object that refers to an object that will start a
+	 * git fetch and then display the 'choose' revision form
+	 */
+	var fetchButton;
+
+
+	/**
+	 * this is a jQuery object that refers the element where the form should be
```
mateusz: this description could be a bit more... descriptive :-)

code/control/DNRoot.php
stojg: Yeah, it was but I have no code to integrate against at this point

PR #298, opened 2015-08-27
First attempt at integrating the DeploymentStrategy with backends.

mateusz:


code/backends/CapistranoDeploymentBackend.php
```diff
@@ -109,7 +101,7 @@ public function deploy(
 		}
 
 		// Check if maintenance page should be removed
-		if(!$leaveMaintenancePage) {
+		if(!empty($options['leaveMaintenancePage'])) {
```
halkyon: Should be `empty()` not `!empty()`

PR #304, opened 2015-08-30
Use a compiled version of react jsx templates

stojg:
Install with npm install and run with `npm run jsx` in this module folder


stojg:
will fix the conflicts

PR #305, opened 2015-08-30
Remove JSON_PRETTY_PRINT

stojg:
Pretty print is a PHP 5.4 "feature"

PR #306, opened 2015-08-30
Drop PHP 5.3 support

stojg:


stojg:
Alternative solution is to drop 5.3 support https://github.com/silverstripe/deploynaut/pull/305

PR #311, opened 2015-08-31
FIX site sha not being passed through correctly.

micmania1:


stojg:
Fixed via https://github.com/silverstripe/deploynaut/pull/301

PR #319, opened 2015-09-01
Adding a proper label to the force full deploy field

stojg:


stojg:
label

stojg:
This PR also switches from using npm run to compile the react and sass into plain js and css to be build by a gulp build script

PR #325, opened 2015-09-04
js build pipeline

stojg:
Javascript from the jsx folder are now concatenated, modularised and minified


stojg:
please hold merging this will I investigate a regression

PR #338, opened 2015-09-08
NEW: users can "star" projects

stojg:
Allow users to star/unstar projects so that the most often used projects can be found quickly

### DIFF
--- a/code/control/DNRoot.php
+++ b/code/control/DNRoot.php
@@ -50,6 +50,7 @@ class DNRoot extends Controller implements PermissionProvider, TemplateGlobalPro
 	 */
 	public static $allowed_actions = array(
 		'projects',
+		'nav',
 		'update',
 		'project',
 		'toggleprojectstar',
@@ -260,6 +261,23 @@ class DNRoot extends Controller implements PermissionProvider, TemplateGlobalPro
 		return $this->customise(array(
 			'Title' => 'Projects',
 		))->render();
+	}
+
+	/**
+	 *
+	 * @param SS_HTTPRequest $request
+	 * @return \SS_HTTPResponse
+	 */
+	public function nav(SS_HTTPRequest $request) {
+		return $this->renderWith('Nav');
+	}
+
+	/**
+	 * Return a link to the navigation template used for AJAX requests.
+	 * @return string
+	 */
+	public function NavLink() {
+		return Controller::join_links(Director::absoluteBaseURL(), 'naut', 'nav');
 	}
 
 	/**
@@ -888,7 +906,10 @@ class DNRoot extends Controller implements PermissionProvider, TemplateGlobalPro
 		$currentProject = $this->getCurrentProject();
 
 		$projects = $this->getStarredProjects();
-		if($projects->count() == 0) {
+		if($projects->count() > 0) {
+			// user has starred projects, show as many as they want.
+			$limit = -1;
+		} else {
 			$projects = $this->DNProjectList();
 		}
 
--- a/code/control/DNRoot.php
+++ b/code/control/DNRoot.php
@@ -52,6 +52,7 @@ class DNRoot extends Controller implements PermissionProvider, TemplateGlobalPro
 		'projects',
 		'update',
 		'project',
+		'toggleprojectstar',
 		'branch',
 		'environment',
 		'abortpipeline',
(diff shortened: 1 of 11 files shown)


stojg:
image

Unfortunately the star/unstar click requires a page reload due to the Stack list on the left. If we want that we need to reload that panel.

stojg:
@clarkepaul any feedback is highly appriciated

code/control/DNRoot.php
```diff
@@ -630,6 +632,33 @@ public function project(SS_HTTPRequest $request) {
...
+
+		$member = Member::currentUser();
+		if($member === null) {
+			return $this->project404Response();
+		}
+		$favProject = $member->StarredProjects()
+			->filter('DNProjectID', $project->ID)
+			->first();
+
+		if($favProject) {
```
halkyon: It may be worth noting that this is actually a "togglestarproject", it doesn't just enable the star?

code/control/DNRoot.php
```diff
@@ -823,6 +852,28 @@ public function DNProjectList() {
...
+		$member = Member::currentUser();
+		if(!$member) {
+			return new ArrayList();
+		}
+
+		$favProjects = $member->StarredProjects();
+
+		$list = new ArrayList();
+		foreach($favProjects as $project) {
+			if($project->CanView($member)) {
```
halkyon: I think this should be `canView`, not `CanView`. PHP is stupid and let's you do whatever case you want.

code/control/DNRoot.php
```diff
@@ -823,6 +852,28 @@ public function DNProjectList() {
 	}
 
 	/**
+	 * Provide a list of all projects.
+	 *
+	 * @return SS_List
+	 */
+	public function getStarredProjects() {
+		$member = Member::currentUser();
+		if(!$member) {
```
halkyon: Should this check be `$member === null` like in the `starproject` controller action?

code/control/DNRoot.php
```diff
@@ -834,7 +885,13 @@ public function Navigation($limit = 5) {
 
 		$currentProject = $this->getCurrentProject();
 
-		$projects = $this->DNProjectList();
+		$projects = $this->getStarredProjects();
+		if($projects->Count() == 0) {
```
halkyon: Should be `count()`, again PHP allowing any case.

code/model/DNProject.php
```diff
@@ -439,6 +440,28 @@ public function Link($action='') {
...
+	/**
+	 * @return bool
+	 */
+	public function IsStarred() {
+		$member = Member::currentUser();
+		if($member === null) {
+			return false;
+		}
+		$favourited = $this->StarredBy()->filter('MemberID', $member->ID);
+		if($favourited->Count() == 0) {
```
halkyon: Should be `count()` here as well

code/control/DNRoot.php
stojg: good suggestion, renamed the method and added a small description on it

stojg:
@halkyon thanks for the review / suggestions

halkyon:
@stojg happy with this from the code perspective. Did you want to wait for @clarkepaul to review as well before we merge?

halkyon:
@stojg one last thing: did you check if this works properly with lots of projects, i.e. doesn't introduce performance regressions?

stojg:
I think paul and someone else should have a look to ensure it's not getting more confusing, for example the nav list doesn't indicate any more that it only displays starred projects. Is that confusing?

And yes, I've checked it with ~40 projects, no perceivable performance impact IMO

stojg:
Feedback: 
 - Left Project list should reload on star/unstar without page reload (ajax)
 - a title or popup on the link that explains what will happen when the star is clicked.

dhensby:
thanks the lawd for this addition :)

dhensby:
I think this looks good - I would say that the `starproject` method should probably check the `canView` of the `$project` before adding it to the many_many, otherwise it might be possible for a malicious user to infer a project url by testing urls and not getting a 404 back.

halkyon:
@dhensby yep, it does that through `DNProjectList` when it calls `getCurrentProject()`

mateusz:
Made only a quick once-over, Stig & Sean have already discussed this extensively.

PR #339, opened 2015-09-08
Adding a basic search filter for project lists that contains more than 5 projects

stojg:


stojg:
Search place holder:

image

During searching

image

stojg:
It will autofocus on the search field when the page have loaded

PR #346, opened 2015-09-09
FIX Starred menu doesn't show correct when selecting non-starred proj…

micmania1:
…ect.


code/control/DNRoot.php
```diff
@@ -906,11 +906,10 @@ public function Navigation($limit = 5) {
...
 
 		$projects = $this->getStarredProjects();
-		if($projects->count() > 0) {
-			// user has starred projects, show as many as they want.
-			$limit = -1;
-		} else {
+		if($projects->count() < 1) {
 			$projects = $this->DNProjectList();
+		} else {
+			$limit -1;
```
halkyon: Shouldn't this be `$limit = -1`?

PR #348, opened 2015-09-09
Ensure that fetching git_revisions adds loading indicator.

halkyon:
This process can take a while on larger repos, e.g. 3-4 seconds.
Rather than show an empty form that looks broken,
show a loading indicator.


jsx/deploy.jsx
```diff
@@ -213,6 +211,10 @@ var deploy = (function (events, classNames) {
 			this.setState({selectedTab: id});
 		},
 		render: function () {
+			if(this.state.loading) {
+				return <LoadingDeployForm message="Loading&hellip;" />
```
micmania1: I think this needs to be wrapped in braces - at least to remain consistent.

PR #351, opened 2015-09-10
Adding compare link to "Code revision" for easy access to diff.

halkyon:
screen shot 2015-09-10 at 12 26 18 pm


clarkepaul:
Wouldn't the arrow just go one way?

PR #354, opened 2015-09-10
Replace eye icon with "View diff"

halkyon:
screen shot 2015-09-10 at 3 17 14 pm


sminnee:
Does anyone thing that "View code changes" or "View changes" would be better?

PR #355, opened 2015-09-10
Show currently deployed version in more clarity and consistency.

halkyon:
Fixing template issue where "Committer" doesn't ever exist, and never
shows the commit message. Add a header "Currently deployed" to
make it a bit more consistent with the deploy screen.


micmania1:
The commit message is set to show after the SHA at the moment - I don't know why it doesn't.

halkyon:
@micmania1 oh really, so it's a bug the commit message doesn't show up :D i'll have a look

halkyon:
@micmania1 Fixed. Removed the "Committer" if blocks in the template, it now exposes the commit message.

halkyon:
Here's what it looks like with this change:

screen shot 2015-09-10 at 4 24 44 pm

sminnee:
i like that we're including all that content, but seems a bit bunched up to me. I'd probably have a clearer indication that the first line is *what* was committed and the second line was *when* it was committed.

PR #356, opened 2015-09-10
Updated resque script to check for running tasks before restarting

micmania1:


.scripts/deploynaut-php-resque
```diff
@@ -79,6 +80,27 @@ do_stop()
 	return "$RETVAL"
 }
 
+#
+# Does some checks before stopping to ensure no tasks are running.
+# This will force a loop until no tasks are running.
+#
+pre_stop_check()
```
mateusz: This still allows the possibility for someone to start a job before you manage to kill off the process.

It seems it would be much safer to use signals - [check out the Resque's QUIT signal handler](https://github.com/chrisboulton/php-resque/blob/master/lib/Resque/Worker.php#L343) . The `start-stop-daemon` in do_stop above is already sending signals, it's just it's using the wrong ones :-)

.scripts/deploynaut-php-resque
```diff
@@ -79,6 +80,27 @@ do_stop()
...
+
+	# if egrep finds something it means a task is working or pending. In this it will
+	# return a 0 exit code. If we get a 0 exit code, we need to prevent the service from stopping.
+	if [ $? -eq 0 ]
+	then
+		echo "Jobs are running. Unable to stop service."
+		echo "Sleeping for 10 seconds."
+		sleep 10
+		pre_stop_check
+	fi
```
mateusz: someone can sneak in deployment between here and do_stop, potentially. Can we just reduce the start-stop-daemon delay to make sure the stray deploy doesn't get too far before it gets killed?

.scripts/deploynaut-php-resque
```diff
@@ -25,6 +25,7 @@ PIDFILE=/var/run/$NAME.pid
 SCRIPTNAME=/etc/init.d/$NAME
 USERNAME=www-data
 LOGFILE=/var/log/$NAME
+#STATUSFILE=$ROOT/resque_status.sh
```
mateusz: And that shouldn't be commented out I think, because you are doing a `[ -f "$STATUSFILE" ]` check below anyway. Noone will remember to comment this out before installing.

.scripts/deploynaut-php-resque
micmania1: If a deployment is started within the 10 second delay, this script would not be stopped until that deployment had also finished. If a deployment was started in the millisecond between pre_stop_check and do_stop, I doubt it would get very far :)

PR #357, opened 2015-09-10
BUG Fixing duplicate tags missing, and ordering of tags.

halkyon:
If multiple tags point to the same revision, these get skipped due to
internally storing revisions by SHA.

Also, tags would show in the reverse order, this changes it so newer
tags appear first in tag listings in the deploy form.

### DIFF
--- a/code/model/DNReferenceList.php
+++ b/code/model/DNReferenceList.php
@@ -141,13 +141,13 @@ class DNReferenceList extends ArrayList {
 			}
 
 			if($this->getTags) {
-				$builds[$reference->getCommitHash()] = DNTag::create($reference, $this->project, $this->data);
-
+				$builds[] = DNTag::create($reference, $this->project, $this->data);
 			} else {
 				$name = $this->reference ? $this->reference->getName() : '';
-				$builds[$reference->getHash()] = DNCommit::create($reference, $this->project, $this->data, $name);
+				$builds[] = DNCommit::create($reference, $this->project, $this->data, $name);
 			}
 		}
+
 		return $builds;
 	}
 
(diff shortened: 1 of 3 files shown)


code/model/DNReferenceList.php
```diff
@@ -141,13 +141,18 @@ protected function getReferences() {
 			}
 
 			if($this->getTags) {
-				$builds[$reference->getCommitHash()] = DNTag::create($reference, $this->project, $this->data);
-
+				$builds[] = DNTag::create($reference, $this->project, $this->data);
 			} else {
 				$name = $this->reference ? $this->reference->getName() : '';
-				$builds[$reference->getHash()] = DNCommit::create($reference, $this->project, $this->data, $name);
+				$builds[] = DNCommit::create($reference, $this->project, $this->data, $name);
```
mateusz: Can you check if commit deduplication in the "redeploy" section still works?

code/model/DNReferenceList.php
```diff
@@ -141,13 +141,18 @@ protected function getReferences() {
...
 			}
 		}
+
+		// reverse the array so tags show with latest first.
+		if($this->getTags) {
+			return array_reverse($builds);
```
mateusz: Should commits not also be sorted in the same way if the tags are? Should be consistent I think.

halkyon: Probably. I'll fix it.

code/model/DNReferenceList.php
```diff
@@ -141,13 +141,18 @@ protected function getReferences() {
 			}
 
 			if($this->getTags) {
-				$builds[$reference->getCommitHash()] = DNTag::create($reference, $this->project, $this->data);
```
mateusz: Actually, why can't we just use the tag name as a hash key here? You wouldn't then need to loop through everything down below.

halkyon: hmm, ok. I guess that makes sense. I'll have a look.

PR #359, opened 2015-09-11
Remove the need to "Verify deployment" button.

halkyon:
Having this button just for admins confuses things, since the end users
don't get this button. It's not necessary, as you can just toggle the
"Force full deployment" checkbox before deploying.


mateusz:
It's there for a reason - its operation is based on the fact that you need to feed it into the planning phase. Now that you've removed the verify button it's state will usually not make it to the backend (nor to the confirmation step) so you will not be able to force full deploy anymore.

PR #364, opened 2015-09-21
Move create form

micmania1:


code/control/DNRoot.php
```diff
@@ -918,11 +963,13 @@ public function doCreateEnvironment($data, Form $form) {
...
 
+		// Set the environment type so we know what we're creating.
+		$data['EnvironmentType'] = $project->AllowedEnvironmentType;
+
 		$job = DNCreateEnvironment::create();
 		$environment = DNEnvironment::create();
 		$environment->ProjectID = $project->ID;
-		$environment->Usage = ''; // todo fill out proper usage here
-		$environment->Name = $project->Cluster . '-' . $project->Name . '-' . $data['Usage'] . '-';
+		$environment->Usage = $datap['Usage'];
```
halkyon: Should be `$data`

code/model/DNProject.php
```diff
@@ -286,6 +287,20 @@ public function canDownloadArchive($member = null) {
 	}
 
 	/**
+	 * This is a permission check for the front-end only.
```
mateusz: This seems to be not untrue? DNRoot::doCreateEnvironment uses that check for backend processing.

code/backends/EnvironmentCreateBackend.php
```diff
@@ -0,0 +1,28 @@
...
+
+interface EnvironmentCreateBackend {
+
+	/**
+	 * @param DNCreateEnvironment $environmet
+	 *
+	 * @return true
+	 * @throws Exception
+	 */
+	public function createEnvironment(DNCreateEnvironment $environment);
```
mateusz: Better call that $environment a $job - it's not a ref to the DNEnvironment.

code/backends/EnvironmentCreateBackend.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
+interface EnvironmentCreateBackend {
+
+	/**
+	 * @param DNCreateEnvironment $environmet
```
mateusz: Typo, environmet

code/model/jobs/DNCreateEnvironment.php
```diff
@@ -132,4 +137,24 @@ public function start() {
...
 
+	public function createEnvironment() {
+		$backend = $this->getBackend();
+		if($backend) {
+			return $backend->createEnvironment($this);
+		}
+		throw new Exception("Unable to find backend.");
+	}
+
+	public function getBackend() {
```
mateusz: This points to the fact that the backend functions should live in the Project backend, not in the Environment backend ;-)

code/model/DNProject.php
```diff
@@ -800,4 +841,22 @@ public function getRepositoryInterface() {
 	protected function getProjectFolderPath() {
 		return $this->DNData()->getEnvironmentDir().'/'.$this->Name;
 	}
+
+	/**
+	 * Create an empty environment based on AllowedEnvironmentType.
+	 *
+	 * @return DNEnvironment
+	 */
+	public function createEmptyEnvironment() {
```
mateusz: Where is this used?

code/control/DNRoot.php
```diff
@@ -918,11 +963,13 @@ public function doCreateEnvironment($data, Form $form) {
...
 		$environment->write();
 
 		$job->Data = serialize($data);
```
mateusz: We'll need to validate that form input when we roll it out to users... otherwise people can switch roles and profiles... Maybe for now put a big TODO here? :-)

code/model/DNProject.php
micmania1: By "front-end" I mean outside of the admin section of the site. The admin section still uses the normal canCreate().

PR #379, opened 2015-09-24
NEW Add extra information on full deploys.

mateusz:


code/model/jobs/DNDeployment.php
```diff
@@ -197,6 +197,66 @@ public function getTags() {
...
+			default: return 'info';
+		}
+	}
+
+	/**
+	 * Convert status to a friendly form, shown on the frontend.
+	 *
+	 * @return string
+	 */
+	public function getStatusFriendly() {
```
halkyon: This isn't needed, as you can just use `ResqueStatus` in the template which remaps "Started" to "Running" as well.

mateusz: I removed the code because it doesn't belong in this commit :P ready for re-review.

PR #389, opened 2015-09-28
Compare URL requires 2 valid 40 character SHAs. Check this before sho…

halkyon:
…wing.

### DIFF
--- a/code/control/DNRoot.php
+++ b/code/control/DNRoot.php
@@ -1370,10 +1370,11 @@ class DNRoot extends Controller implements PermissionProvider, TemplateGlobalPro
 		);
 		$data = $strategy->toArray();
 
-		// Add in a URL for comparing from->to code changes.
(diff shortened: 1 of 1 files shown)


code/control/DNRoot.php
```diff
@@ -1370,10 +1370,11 @@ public function deploySummary(SS_HTTPRequest $request) {
 		);
 		$data = $strategy->toArray();
 
-		// Add in a URL for comparing from->to code changes.
+		// Add in a URL for comparing from->to code changes. Ensure that we have
+		// two proper 40 character SHAs, otherwise we can't show the compare link.
```
mateusz: Is 1.2..1.3 not a valid comparison? Will from & to always be 40-char sha's? 

Why does this fail anyway, what's the trigger?

halkyon: No, they have to be full SHAs. One of the from values can be "-" (which is the case for new environments)

PR #390, opened 2015-09-29
API Expand the project-centered "allow" API.

mateusz:


code/model/DNEnvironment.php
```diff
@@ -392,8 +392,10 @@ public function canDeploy($member = null) {
 		}
 		// Must be logged in to check permissions
 
-		if(Permission::checkMember($member, 'ADMIN')) {
-			return true;
+		if ($this->Usage==='Production' || $this->Usage==='Unspecified') {
```
stojg: Can we define a constant for the magic string 'Production', 'UAT', and 'Unspecified' for the `Usage` field?

mateusz: I'm not sure about that - this comes from a DB Enum structure, so is itself fairly constant. We don't define consts for any other state-machine enums in deployments either so it would also be inconsistent.

stojg: fair enough, Just worried that a misspelling of 'Production' would break this quite easily

PR #392, opened 2015-09-30
NEW Introduce extensible ambient menu (i.e. the one with logout button)

mateusz:


code/control/DNRoot.php
```diff
@@ -2322,4 +2322,26 @@ protected function getCustomisedViewSection($sectionName, $title = '') {
...
+	 *
+	 * @return ArrayList
+	 */
+	public function AmbientMenu() {
+		$list = new ArrayList();
+
+		if (Member::currentUserID()) {
+			$list->push(new ArrayData(array(
+				'FaIcon' => 'sign-out',
+				'Link' => 'Security/logout',
```
micmania1: Can you use Security::logout_url()?

mateusz: This is not an API call?

micmania1: oh, sorry. That's in 3.2 :D

PR #394, opened 2015-10-01
FIX popover wasn't wide enough to fit most content.

micmania1:


mateusz:
Needs resubmission against master :-|

PR #421, opened 2015-10-14
Added the ability to add a project on the front end for administrators.

micmania1:
This provides basic functionality for creating a new project in the front-end with all necessary fields.

What it doesn't do:
* Editing viewer groups via the front-end
* Create environments

create-project-2

### DIFF
--- a/code/control/DNRoot.php
+++ b/code/control/DNRoot.php
@@ -50,7 +50,7 @@ class DNRoot extends Controller implements PermissionProvider, TemplateGlobalPro
 	/**
 	 * @var array
 	 */
-	public static $allowed_actions = array(
+	private static $allowed_actions = array(
 		'projects',
 		'nav',
 		'update',
@@ -88,12 +88,14 @@ class DNRoot extends Controller implements PermissionProvider, TemplateGlobalPro
 		'gitRevisions',
 		'deploySummary',
 		'startDeploy',
+		'createproject',
+		'CreateProjectForm',
 	);
 
 	/**
 	 * URL handlers pretending that we have a deep URL structure.
 	 */
-	public static $url_handlers = array(
+	private static $url_handlers = array(
 		'project/$Project/environment/$Environment/DeployForm' => 'getDeployForm',
 		'project/$Project/createsnapshot/DataTransferForm' => 'getDataTransferForm',
 		'project/$Project/DataTransferForm' => 'getDataTransferForm',
@@ -2337,6 +2339,103 @@ class DNRoot extends Controller implements PermissionProvider, TemplateGlobalPro
 
 		$this->extend('updateAmbientMenu', $list);
 		return $list;
+	}
+
+	/**
+	 * Create project action.
+	 *
+	 * @return SS_HTTPResponse
+	 */
+	public function createproject(SS_HTTPRequest $request) {
+		if($this->canCreateProjects()) {
+			return $this->render(['CurrentTitle' => 'Create Stack']);
+		}
+		return $this->httpError(403);
+	}
+
+	/**
+	 * Checks whether the user can create a project.
+	 *
+	 * @return bool
+	 */
+	protected function canCreateProjects($member = null) {
+		if(!$member) $member = Member::currentUser();
+		if(!$member) return false;
+
+		return Permission::checkMember($member, 'ADMIN');
+	}
+		
+	/**
+	 * @return Form
+	 */
+	public function CreateProjectForm() {
+		$form = Form::create(
+			$this, 
+			__FUNCTION__, 
+			$this->getCreateProjectFormFields(), 
+			$this->getCreateProjectFormActions(),
+			new RequiredFields('Name', 'CVSPath')
+		);
+		$this->extend('updateCreateProjectForm', $form);
+		return $form;
+	}
+
+	/**
+	 * @return FieldList
+	 */
+	protected function getCreateProjectFormFields() {
+		$fields = FieldList::create();
+		$fields->merge([
+			TextField::create('Name', 'Title'),
+			TextField::create('CVSPath', 'Git URL'),
+			TextField::create('Client', 'Client'),
+		]);
+		$this->extend('updateCreateProjectFormFields', $fields);
+		return $fields;
+	}
+
+	/**
+	 * @return FieldList
+	 */
+	protected function getCreateProjectFormActions() {
+		$fields = FieldList::create(
+			FormAction::create('doCreateProject', 'Create Stack')
+		);
+		$this->extend('updateCreateProjectFormActions', $fields);
+		return $fields;
+	}
+
+	/**
+	 * Does the actual project creation.
+	 *
+	 * @param $data array
+	 * @param $form Form
+	 *
+	 * @return SS_HTTPResponse
+	 */
+	public function doCreateProject($data, $form) {
+		$form->loadDataFrom($data);
+		$project = DNProject::create();
+
+		$form->saveInto($project);
+		$this->extend('onBeforeCreateProject', $project, $data, $form);
+		try {
+			if($project->write() > 0) {
+				$this->extend('onAfterCreateProject', $project, $data, $form);
+
+				// If an extension hasn't redirected us, we'll redirect to the project.
+				if(!$this->redirectedTo()) {
+					return $this->redirect($project->Link());
+				} else {
+					return $this->response;
+				}
+			} else {
+				$form->sessionMessage('Unable to write the stack to the database.', 'bad');
+			}
+		} catch (ValidationException $e) {
+			$form->sessionMessage($e->getMessage(), 'bad');
+		}
+		return $this->redirectBack();
 	}
 
 }
--- a/code/jobs/CreateEnvJob.php
+++ b/code/jobs/CreateEnvJob.php
@@ -29,35 +29,8 @@ class CreateEnvJob {
 	 */
 	public function perform() {
 		echo "[-] CreateEnvJob starting" . PHP_EOL;
-		$log = new DeploynautLogFile($this->args['logfile']);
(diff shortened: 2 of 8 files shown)


templates/Layout/DNRoot_createproject.ss
```diff
@@ -0,0 +1,49 @@
...
+
+<div class="content">
+
+	<div class="row">
+		<div class="col-md-12 environment-details">
+			<a href="$CurrentProject.Link"><i class="fa fa-long-arrow-left"></i> Back to overview</a>
+		</div>
+	</div>
+
+	<% if $CreateProjectProgress %>
```
halkyon: Where is `CreateProjectProgress` defined?

micmania1: Nowhere - I had planned to but never did. I'll remove it first.

code/control/DNRoot.php
```diff
@@ -2339,4 +2346,99 @@ public function AmbientMenu() {
...
+	}
+
+	/**
+	 * Checks whether the user can create a project.
+	 *
+	 * @return bool
+	 */
+	protected function canCreateProjects() {
+		$project = Injector::inst()->get('DNProject');
+		return $project->allowed(self::ALLOW_CREATE_PROJECT);
```
mateusz: "allowed" class of permission checkers are bound to a project. Can you rename this permission to not use this keyword and use normal Permission::check?

code/control/DNRoot.php
```diff
@@ -88,12 +89,14 @@ class DNRoot extends Controller implements PermissionProvider, TemplateGlobalPro
 		'gitRevisions',
 		'deploySummary',
 		'startDeploy',
+		'createproject',
+		'CreateProjectForm',
```
mateusz: Why not use ModelAdmin as Sam suggested if it's only for admins?

templates/Layout/DNRoot_projects.ss
```diff
@@ -6,6 +6,14 @@
 	</div>
 </div>
 <div class="content">
+
+	<% if $canCreateProjects %>
+		<a href="{$Link}createproject" class="btn btn-primary pull-right">
+			<i class="fa fa-plus"></i>
+			Create Stack
```
mateusz: Sounds misnamed.

code/model/DNProject.php
```diff
@@ -969,5 +969,30 @@ public function allowedAny($codes, $member = null) {
 		return ($hits>0 ? true : false);
 	}
 
+	/**
+	 * @return ValidationResult
+	 */
+	protected function validate() {
+		$validation = parent::validate();
+		if($validation->valid()) {
+			if(empty($this->Name)) {
```
mateusz: Does project name need to be unique?

templates/Layout/DNRoot_projects.ss
mateusz: Oh project=stack. Blergh. Can we rename it all to stack? :-)

code/control/DNRoot.php
```diff
@@ -2339,4 +2346,99 @@ public function AmbientMenu() {
...
+				$this->extend('onAfterCreateProject', $project, $data, $form);
+
+				// If an extension hasn't redirected us, we'll redirect to the project.
+				if(!$this->redirectedTo()) {
+					return $this->redirect($project->Link());
+				} else {
+					return $this->response;
+				}
+			} else {
+				$form->sessionMessage('Unable to write the project to the database.', 'bad');
```
mateusz: "Unable to write the stack..."

code/model/DNProject.php
mateusz: Ah you have it down there, cool.

code/model/DNProject.php
```diff
@@ -969,5 +969,30 @@ public function allowedAny($codes, $member = null) {
...
+			if(empty($this->CVSPath)) {
+				return $validation->error('You must provide a repository URL.');
+			}
+
+			$existing = DNProject::get()->filter('Name', $this->Name);
+			if($this->ID) {
+				$existing = $existing->exclude('ID', $this->ID);
+			}
+			if($existing->count() > 0) {
+				return $validation->error('A project already exists with that name.');
```
mateusz: "A stack..."

code/jobs/CreateEnvJob.php
```diff
@@ -28,36 +28,8 @@ public function tearDown() {
 	 *
 	 */
 	public function perform() {
-		echo "[-] CreateEnvJob starting" . PHP_EOL;
```
mateusz: Why remove debug output? We should have more of it if anything :-)

code/control/DNRoot.php
```diff
@@ -2339,4 +2346,99 @@ public function AmbientMenu() {
...
+	 * @return SS_HTTPResponse
+	 */
+	public function doCreateProject($data, $form) {
+		$form->loadDataFrom($data);
+		$project = DNProject::create();
+
+		$form->saveInto($project);
+		$this->extend('onBeforeCreateProject', $project, $data, $form);
+		try {
+			if($project->write() > 0) {
```
mateusz: Can we generate some metrics, like with other Create things?

code/jobs/CreateEnvJob.php
```diff
@@ -28,36 +28,8 @@ public function tearDown() {
...
-					'Status' => array('Queued', 'Started'),
-					'Created:GreaterThan' => strtotime('-30 minutes')
-				))
-				->exclude('ID', $this->args['createID']);
-
-
-			if($running->count()) {
-				$existing = $running->first();
-				$log->write(sprintf(
-					'[-] Error: another create job is in progress (started at %s by %s)',
```
mateusz: I assume this is being handled elsewhere now?

code/control/DNRoot.php
micmania1: "The UI is good enough for customers" - acceptance criteria. It will initially be used by Ops but will be rolled out to normal users and be part of the trial system.

code/jobs/CreateEnvJob.php
micmania1: Nope, you should be able to create multiple envs at once (see RF fix). Use case is where the customer creates the new stack and both envs need to be created. Given this will usually be their first interaction with platform, I wanted the creation to be quick (half the time).

code/control/DNRoot.php
micmania1: There are hooks where metrics can be tied in.

code/control/DNRoot.php
mateusz: Ah ok thanks, I thought we are still scaffolding. Thanks for explaining the driver.

code/jobs/CreateEnvJob.php
mateusz: How will you ensure deploynaut is not starved out of workers?

code/model/DNProject.php
```diff
@@ -969,5 +969,30 @@ public function allowedAny($codes, $member = null) {
...
 	}
 
+	/**
+	 * @return ValidationResult
+	 */
+	protected function validate() {
+		$validation = parent::validate();
+		if($validation->valid()) {
+			if(empty($this->Name)) {
+				return $validation->error('The project must have a name.');
```
mateusz: "The stack..."

templates/Layout/DNRoot_projects.ss
```diff
@@ -6,6 +6,14 @@
...
+		</a>
+	<% end_if %>
+
 	<%-- Dont show the search/filter if there are less than 5 projects in the list --%>
 	<% if $DNProjectList.count > 5 %>
 	<input type="search" class="table-filter form-control" data-table="table-projects" placeholder="Search for project">
```
mateusz: "Search for stack" :-)

code/control/DNRoot.php
```diff
@@ -2339,4 +2341,101 @@ public function AmbientMenu() {
...
+		$this->extend('updateCreateProjectForm', $form);
+		return $form;
+	}
+
+	/**
+	 * @return FieldList
+	 */
+	protected function getCreateProjectFormFields() {
+		$fields = FieldList::create();
+		$fields->merge([
```
mateusz: If we are going for the A/C then we are missing "Stack codename" and "Cluster".

code/control/DNRoot.php
mateusz: Put a task in if you don't want to do it now? :-)

code/control/DNRoot.php
```diff
@@ -2339,4 +2341,101 @@ public function AmbientMenu() {
...
+
+	/**
+	 * Does the actual project creation.
+	 *
+	 * @param $data array
+	 * @param $form Form
+	 *
+	 * @return SS_HTTPResponse
+	 */
+	public function doCreateProject($data, $form) {
```
mateusz: Also from A/C, how about adding the creator as Stack Manager?

code/control/DNRoot.php
micmania1: Do we log any metrics in this module? Seems a strange place to be logging metrics.

code/jobs/CreateEnvJob.php
micmania1: How can we ever? If 10 different people create an environment at the same time then we'd have the same issue. This only restricts on a per-project basis so doesn't solve that problem and doesn't prevent dev's solving this issue elsewhere if its an issue (ie. at the creation form instead of resque worker). There's hooks provided to allow for extra validation.

code/control/DNRoot.php
micmania1: Different module :)

code/control/DNRoot.php
micmania1: Different module :)

micmania1:
Fixed some stuff. I've left comments on others @mateusz

PR #424, opened 2015-10-16
Cross link frontend and backend

mateusz:
screen shot 2015-10-16 at 3 32 45 pm
screen shot 2015-10-16 at 3 32 49 pm

### DIFF
--- a/templates/Includes/ProjectLinks.ss
+++ b/templates/Includes/ProjectLinks.ss
@@ -1,5 +1,13 @@
 <% if $CurrentProject %>
 <ul class="project-links">
+	<% if $HasPerm(ADMIN) %>
+		<li>
+			<i class="fa fa-stethoscope"></i>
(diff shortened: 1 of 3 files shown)


templates/Includes/ProjectLinks.ss
```diff
@@ -1,5 +1,13 @@
 <% if $CurrentProject %>
 <ul class="project-links">
+	<% if $HasPerm(ADMIN) %>
+		<li>
+			<i class="fa fa-stethoscope"></i>
+			<a target="deploynaut-backend" href="admin/naut/DNProject/EditForm/field/DNProject/item/{$CurrentProject.ID}/ItemEditForm/field/Environments/item/{$CurrentEnvironment.ID}/edit">
+				CMS
```
halkyon: Edit in CMS with a cog or pencil icon instead?

mateusz: Updated the text. But I want the stethoscope. Most of the time you'd pop over to the CMS to diagnose problems, and it's OPS only feature anyway.

PR #427, opened 2015-10-21
Added project creation waiting page with deploy key testing

micmania1:


templates/Includes/Nav.ss
```diff
@@ -8,16 +8,20 @@
 
 			<ul class="collapse<% if $IsActive %> in<% end_if %>" id="collapseExample-$Pos">
 				<% with $Project %>
-					<% if $DNEnvironmentList %>
-						<li>
-							<a role="button" href="$Link">Environments</a>
+					<li>
+						<a role="button" href="$Link">Environments</a>
+						<% if IsProjectReady && $DNEnvironmentList %>
```
halkyon: Don't forget to add `$` to the front of template controls ;-)

js/create_project_progress.jsx
```diff
@@ -0,0 +1,83 @@
...
+
+	getInitialState: function() {
+		return {
+			complete: false,
+			checkingStatus: false
+		};
+	},
+
+	componentDidMount: function() {
+		this.checkInterval = window.setInterval(this.handleCheckStatus, 10000);
```
mateusz: We should probably get into a habit of cleaning stuff up - how about you kill off the timer in componentWillUnmount?

code/control/DNRoot.php
```diff
@@ -2438,4 +2440,93 @@ public function doCreateProject($data, $form) {
...
+	 *
+	 * @return SS_HTTPResponse
+	 */
+	public function createprojectprogress(SS_HTTPRequest $request) {
+		$project = $this->getCurrentProject();
+		if(!$project) {
+			return $this->httpError(404);
+		}
+
+		if(!$project->canView()) {
```
halkyon: canView isn't needed as `getCurrentProject()` goes through the process of doing that already via `DNProjectList()`

code/jobs/CloneGitRepo.php
```diff
@@ -52,7 +52,6 @@ public function perform() {
 		}
 
 		fwrite($fh, sprintf('[%s] Cloning repository %s to %s', date('Y-m-d H:i:s'), $repo, $path) . PHP_EOL);
-		echo "[-] CloneGitRepo starting" . PHP_EOL;
```
mateusz: Why remove debug?

micmania1: It echo's out to the page which is an issue because I'm trying to return json data.

mateusz: Oh ok.

mateusz:
Feedback from the demo: if project/env creation errors out, make sure the user is told, so they know to talk to Ops and not wait indefinitely. It could be as simple as checking if there are at least 2 envs, and if less, then showing some error.

js/create_project_progress.jsx
micmania1: This is cleaned up on line 34: https://github.com/silverstripe/deploynaut/pull/427/files#diff-c5e6c9222980acfeb9e5ecc3f982755dR34

micmania1:
@mateusz I've added the failure detection. It will only show failure at the end after all environments have completed or failed.

js/create_project_progress.jsx
mateusz: I meant clearing the timer in the `componentWillUnmount` function. This is just best practice https://facebook.github.io/react/docs/component-specs.html#unmounting-componentwillunmount . My understanding is, it's invalid to work with a component that has been unmounted.

mateusz:
@micmania1 I don't see where the failure is exposed (just looking at the code). I still only see "Your environments are setup and ready to go!" as the only exit from the environment creation screen. Can you point me to the right location?

micmania1:
@mateusz I committed it to the wrong branch :/ Its there now.

js/create_project_progress.jsx
```diff
@@ -0,0 +1,145 @@
...
+				});	   
+
+
+
+				var completeMessage = (
+					<div>
+						<i className="fa fa-exclamation fa-5x text-danger"></i>
+						<h1>Uh Oh!</h1>
+						<p>Something went wrong and we had trouble creating your environments.</p>
+						<p>Below are a list of environments which failed to build:</p>
```
mateusz: Below *is* a list of environments...?

micmania1: I think 'are' is correct here. Second opinion needed :D

halkyon: or "These environments failed to build:"

js/create_project_progress.jsx
micmania1: I tried this but it didn't clear.

mateusz: It didn't clear because your CreateProjectProgress component did not unmount - it's still visible till the end. However, if you ever put that component under another component, and then the state changes in a way that removes it, the timer will keep ticking away in the background, and then fail. So you need it both in success (no need to poll after finishing) AND in unmount (as a best practice).

PR #430, opened 2015-10-22
Move the env creation activity icon to top element. Add to main listing.

mateusz:
test

---

test2

### DIFF
--- a/templates/Includes/Nav.ss
+++ b/templates/Includes/Nav.ss
@@ -3,7 +3,11 @@
 	<% loop $Navigation(5) %>
 		<li class="active">
 			<a class="nav-submenu<% if $IsActive %> open<% end_if %>" role="button" data-toggle="collapse" href="#collapseExample-$Pos" aria-controls="collapseExample-$Pos">
-				<span class="plat-icon icon-stack"></span>$Project.Name<span class="icon-arrow"></span>
+				<% if $Project.RunningEnvironmentCreations.count>0 %>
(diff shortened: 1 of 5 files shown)


templates/Includes/Nav.ss
```diff
@@ -3,7 +3,11 @@
 	<% loop $Navigation(5) %>
 		<li class="active">
 			<a class="nav-submenu<% if $IsActive %> open<% end_if %>" role="button" data-toggle="collapse" href="#collapseExample-$Pos" aria-controls="collapseExample-$Pos">
-				<span class="plat-icon icon-stack"></span>$Project.Name<span class="icon-arrow"></span>
+				<% if $Project.RunningEnvironmentCreations.count>0 %>
+					<span class="plat-icon fa fa-cog fa-spin tooltip-hint" data-original-title="Environment build in progress"></span>$Project.Name<span class="icon-arrow"></span>
```
halkyon: "Environments are currently being built" instead? :)

halkyon:
Can we have it so the icon disappears once the environments are finished, rather than having to refresh the page? We've already got such things like the username updating when someone changes their profile, would make sense to do it in other places too :)

mateusz:
@halkyon that would require a long poll in the background, and these bits of the interface are not reactify, so I'm calling it nice-to-have. The only one that might be worth doing is the one when the "currently building" splash screen transitions to "complete". Added a task.

Otherwise good to merge?

PR #435, opened 2015-10-26
FIX Improved validation on project name.

micmania1:


micmania1:
This doesn't fix the entire problem.

PR #436, opened 2015-10-26
BUG Fix AJAX refresh of the side menu.

mateusz:
- AJAX fetch of the nav was not aware of current project, ended up
  always collapsing all menus. Added a URL param to fix it.
- remove cogwheel when project creation completes


code/control/DNRoot.php
```diff
@@ -2361,6 +2364,7 @@ public function createproject(SS_HTTPRequest $request) {
 	 * @return bool
 	 */
 	protected function canCreateProjects($member = null) {
+		return true;
```
halkyon: huhhhh

PR #439, opened 2015-10-27
Remove client field from project.

micmania1:


code/control/DNRoot.php
```diff
@@ -2390,11 +2390,10 @@ public function CreateProjectForm() {
 	 */
 	protected function getCreateProjectFormFields() {
 		$fields = FieldList::create();
-		$fields->merge([
+		$fields->merge(array(
```
halkyon: No love for the `[]`? :(

micmania1: PHP 5.3 still :(

halkyon: nah, we've bumped to 5.4 minimum

PR #442, opened 2015-10-29
Tweak ul margin within alert box so it's consistent with paragraphs.

halkyon:


stojg:
A screenshot would be helpful :)

PR #453, opened 2015-11-02
Make the width of the top nav items smaller for smaller viewscreens

stojg:
It's not a giant fix, but tightens it up a little bit more

## Before - xtra small

before-xsmall

## After  - xtra small

after-xsmall

## Before - large
before-large

## After - large
after-large


stojg:
Please refrain from pushing the green merge button, a wild bug has appeared

PR #469, opened 2015-11-09
Cleanup CreateEnvironment after environment is deleted.

halkyon:


code/model/DNEnvironment.php
```diff
@@ -1160,6 +1160,11 @@ public function onAfterDelete() {
 		if($this->config()->get('allow_web_editing') && $this->envFileExists()) {
 			unlink($this->getConfigFilename());
 		}
+
+		$create = $this->CreateEnvironment();
```
stojg: it's a bit funny name for a relation.. Looks like call to create an environment.. :P

halkyon: yeah... we'll probably refactor the whole create env thing later anyway. Good to merge?

PR #481, opened 2015-11-16
Adding a event pub/sub system

stojg:


code/events/GenericEvent.php
```diff
@@ -0,0 +1,90 @@
...
+ * ]));
+ *
+ * Then a listener can subscribe the event like this (_config.php):
+ *
+ * $emitter = Injector::inst()->get('EventEmitter');
+ * $emitter->addListener('event.name', new DomainListener());
+ *
+ * See more on how to use the event system here: http://event.thephpleague.com/2.0/
+ */
+class GenericEvent extends AbstractEvent {
```
micmania1: If you're adding more public API here, then you'll want to create a sub interface of League\Event\EventEmitter and implement it here so that any subclasses can be used without fear of API being different.

code/events/GenericEvent.php
```diff
@@ -0,0 +1,90 @@
...
+	protected $data = [];
+
+	/**
+	 * @param string $name - event name
+	 * @param array $data
+	 *
+	 * @return GenericEvent
+	 */
+	public static function create($name, array $data) {
+		return new GenericEvent($name, $data);
```
micmania1: Use Injector->createWithArgs()? I'll merge with or without so i'll leave it up to you.

PR #482, opened 2015-11-17
Show all deployment details from the strategy.

halkyon:
screen shot 2015-11-18 at 12 40 36 pm
screen shot 2015-11-18 at 12 40 44 pm


clarkepaul:
Its getting pretty cluttered, can we use the status as the header for the details? The table data can remain as the same style but move below the status bar. 

pasted_image_18_11_15__2_08_pm

halkyon:
@clarkepaul Updated. New screenshots:

screen shot 2015-11-18 at 3 31 30 pm
screen shot 2015-11-18 at 3 31 36 pm

sminnee:
:+1: from me!

mandrew:
I like the change but why was the URL to the website removed?

halkyon:
@mandrew it's on the environment page, it doesn't need to be shown on deployment as it's not related.

clarkepaul:
Looking better I think - to make a little cleaner (if you have time):
- make the words "view details" font-size smaller.
- the underline extends past the words "View details" because of the space, use margin instead of cursor space.
- the underline might not be actually needed, but not a biggy?

halkyon:
@clarkepaul Done. It's now consistent with the deploy form (in terms of where the git repo fetch text is)

screen shot 2015-11-18 at 3 40 26 pm

mandrew:
@halkyon but you might want to check the website and see if it is ok? ;) 
One option could be making the title a link to the environment url, or making it another link on the righthand side but even that list of links is getting to long for the right column? /cc @clarkepaul

clarkepaul:
Coolio

clarkepaul:
@mandrew yeah I know that those links are way out of control - we think its nearly time for the overview dashboard where some of those links could live.

halkyon:
@mandrew The URLs are already on the stack overview _and_ environment page, why do they need to be everywhere? ;-)

mandrew:
@clarkepaul yeah I guess those links relate to all environments rather then just the environment you are viewing so could be moved into the main Environment holder page.

@halkyon I've never used them though ; ) however I have always clicked on the url once the deployment has finished on the deplyment complete page (well either that or typed it in manually heh).

stojg:
Looks good to me for now. We can tighten up things in general in another story, which would be needed before it goes too far down rabbit hole.

dhensby:
There's an issue with this where the "view details" link is not `e.preventDefault()` on the click and so the browser jumps to the top of the page

PR #484, opened 2015-11-18
Added hooks for permission extensions and made deploynaut first item

micmania1:
This allows to extend the canView permission of DNProject. I've also made it so DNAdmin is navigated to first in the CMS.

### DIFF
--- a/code/model/DNProject.php
+++ b/code/model/DNProject.php
@@ -253,7 +253,29 @@ class DNProject extends DataObject {
 			return true;
 		}
 
-		return $member->inGroups($this->Viewers());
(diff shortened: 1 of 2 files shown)


code/model/DNProject.php
```diff
@@ -253,7 +253,29 @@ public function canView($member = null) {
 			return true;
 		}
 
-		return $member->inGroups($this->Viewers());
+		if($member->inGroups($this->Viewers())) {
+			return true;
+		}
+
+		// Calls extsions
```
halkyon: Looks good, "extensions" is misspelled here.

micmania1:
erm... I'm going to close this.

PR #490, opened 2016-01-06
NEW Add checkbox for skipping the pre-deployment snapshot in pipelines.

mateusz:
## Deployment screen

screen shot 2016-01-07 at 10 11 21 am

## Pipeline-in-progress screen

screen shot 2016-01-07 at 9 58 17 am

### DIFF
--- a/code/control/DeployForm.php
+++ b/code/control/DeployForm.php
@@ -120,6 +120,28 @@ class DeployForm extends Form {
 				$validator = new DeployForm_PipelineValidator();
 			}
 
+			$skipSnapshotTip = <<<END
+				In the unusual occurence of your deployment containing database changes that are
+				not reversible without a snapshot, rollback will not be able to restore your site proprely.
+				The site will remain in maintenance mode - please contact the Service Desk for assistance.
+END;
+
+			$fieldList = new FieldList(array(
+				HeaderField::create('Label1', '1. Choose a commit to deploy', 4),
+				$field,
+				HeaderField::create('Label2', '2. Advanced options', 4),
+				CheckboxField::create(
+					'SkipSnapshot',
+					sprintf(
+						'Skip database snapshot and use existing database for rollbacks '
+						. '(<span class="tooltip-hint" href="#" data-toggle="tooltip" data-placement="bottom" title="%s">'
+						. 'understand the risk'
+						. '</span>)',
+						$skipSnapshotTip
+					)
+				)
+			));
+
 			// Generate actions allowed for this user
 			$actions = new FieldList(
 				FormAction::create('startPipeline', "Begin the release process on " . $environment->Name)
@@ -159,8 +181,10 @@ class DeployForm extends Form {
 						"return confirm('This will start a direct deployment.\\n\\nContinue?');"
 					)
 			);
+
+			$fieldList = new FieldList($field);
 		}
-		parent::__construct($controller, $name, new FieldList($field), $actions, $validator);
+		parent::__construct($controller, $name, $fieldList, $actions, $validator);
 	}
 
 	/**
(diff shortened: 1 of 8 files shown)


tests/DeploymentPipelineStepTest.php
```diff
@@ -102,6 +103,21 @@ public function testSnapshotSkipped() {
...
+	public function testSnapshotSkippedByUser() {
+		$step = $this->getDummyDeployment(false, true);
+		$step->start();
+
+		// Assert not error at startup
+		$this->assertEquals('Started', $step->Status);
+		$this->assertEquals('Deployment', $step->Doing);
+		$this->assertFalse(PipelineTest_MockLog::has_message('TestDeployStep:Snapshot creating snapshot of database'));
+		$snapshot = $step->Pipeline()->PreviousSnapshot();
+		$this->assertFalse($snapshot->exists(), 'No snapshot was created');
```
stojg: shouldn't the message be something like: 'lol, a snapshot was created, it shouldn't have been created?

tests/DeploymentPipelineStepTest.php
```diff
@@ -102,6 +103,21 @@ public function testSnapshotSkipped() {
...
+	 */
+	public function testSnapshotSkippedByUser() {
+		$step = $this->getDummyDeployment(false, true);
+		$step->start();
+
+		// Assert not error at startup
+		$this->assertEquals('Started', $step->Status);
+		$this->assertEquals('Deployment', $step->Doing);
+		$this->assertFalse(PipelineTest_MockLog::has_message('TestDeployStep:Snapshot creating snapshot of database'));
+		$snapshot = $step->Pipeline()->PreviousSnapshot();
```
stojg: Can you assert that a deployment was started as well? Since that is the expected behaviour?

code/control/DeployForm.php
```diff
@@ -120,6 +120,16 @@ public function __construct($controller, $name, DNEnvironment $environment, DNPr
 				$validator = new DeployForm_PipelineValidator();
 			}
 
+			$fieldList = new FieldList(array(
```
stojg: wasn't this only supposed to be shown for super users?

mateusz: Well, good point. https://silverstripe.atlassian.net/browse/CWP-753 doesn't mention anything. Will enquire.

tests/DeploymentPipelineStepTest.php
mateusz: that's done above, with `$this->assertEquals('Deployment', $step->Doing);`

tests/DeploymentPipelineStepTest.php
mateusz: Fixing.

mateusz: Ah, nah, the message is supposed to assert what should have happened. So it's correct.

tests/DeploymentPipelineStepTest.php
stojg: `doing`... oka.

tests/DeploymentPipelineStepTest.php
stojg: the output will be like:

```
There was 1 failure:

1) DeploymentPipelineStepTest::testSnapshotSkippedByUser
No snapshot was created
Failed asserting that true is false.
```

Would that make sense? You are trying to assert that there no snapshot created. Just thinking about what sort of "standard" we should have in the assert messages, since I try to point out why the test failed. no biggie, just interested in your opinions.

stojg: `Snapshot shouldn't have been created` ?

mateusz: Removed both occurences of confusing description. No other asserts in this test use descriptions either, so that's at least consistent.

mateusz:
@stojg added two screenshots of the UI in the PR description.

PR #502, opened 2016-02-04
Added template include to environment page

yarthetpan:


templates/Layout/DNRoot_environment.ss
```diff
@@ -10,22 +10,7 @@
 
 <div class="content">
 
-	<% with $CurrentEnvironment %>
```
mateusz: Why move this bit into the new extension point? That bit in here already contains an extension in fact :-)

PR #518, opened 2016-02-29
Improved detection on failed snapshots to prevent stale 'running' jobs.

micmania1:


_config.php
```diff
@@ -6,3 +6,11 @@
 $exts = Config::inst()->get('File', 'allowed_extensions');
 $exts[] = 'sspak';
 Config::inst()->update('File', 'allowed_extensions', $exts);
+
+// This will ensure jobs can correctly clean themselves up on any type of failure
+Resque_Event::listen('onFailure', function(Exception $exception, Resque_job $job) {
+	$inst = $job->getInstance();
+	if($inst instanceof DeploynautJobInterface) {
+		$inst->onFailure($exception);
```
halkyon: Do we need to pass `$exception` in given we don't use it?

micmania1: Nah, I thought it could be used in the future. eg. The instance class may have a logger attached to it which would log the exception message. I can remove if you don't think its worth having.

halkyon: Now that you say that it might be worth having :) @stojg what do you reckon?

stojg: keep it, as mike was saying, it might have a logger, or should have.

PR #520, opened 2016-02-29
Remove unncessary tear downs/failure setting in jobs.

halkyon:


code/jobs/CreateEnvJob.php
```diff
@@ -15,10 +15,6 @@ public function setUp() {
 		chdir(BASE_PATH);
 	}
 
-	public function onFailure(Exception $exception) {
```
micmania1: Is this because it exists in DeploynautJob?

halkyon: @micmania1 yeah, it's in the parent class so we don't need this duplicate :)

PR #525, opened 2016-03-02
Fixing getChanges to return "empty" values as empty strings for comparison

halkyon:


code/backends/DeploymentStrategy.php
```diff
@@ -160,7 +160,18 @@ public function getChangesModificationNeeded() {
 	 *	)
 	 */
 	public function getChanges() {
-		return $this->changes;
+		// Normalise "empty" values into dashes so comparisons are done properly.
+		// This means there is no diference between an empty string and a null
+		// but "0" is considered to be non-empty.
+		$changes = $this->changes;
```
stojg: how about (string)"false" ? :smiling_imp:

stojg:
So the latest idea (non implemented) is to keep all the values falsish in the deployment strategy and only change empty values in the front-end to dashes. Sounds like a plan @halkyon ?

halkyon:
@stojg yep, I'll write some tests for it too... can only go so far keeping it in your head :)

PR #527, opened 2016-03-04
fetch the deploy strategy from the environment instead of chasing the deploybackend

stojg:


stojg:
I think the DNEnvironments needs to planning the deploys, in collaboration with it's deployment backend. This is a first step to break the direct dependency between DNRoot -> deployment backend.

code/model/DNEnvironment.php
```diff
@@ -210,6 +210,15 @@ public function Backend() {
 		return Injector::inst()->get($backend);
 	}
 
+	/**
+	 * @param SS_HTTPRequest $request
+	 *
+	 * @return DeploymentStrategy
+	 */
+	public function getDeployStrategy(\SS_HTTPRequest $request) {
+		return $this->Backend()->planDeploy($this, $request);;
```
halkyon: DOUBLE COLON

stojg: double tap

mateusz:
As you say, sir!

PR #532, opened 2016-03-17
Ensure that we only setState on a mounted form

stojg:


stojg:
Please don't merge this yet, still investigating some edge cases.

PR #545, opened 2016-04-11
Support readonly items in variable editor and table components

halkyon:


js/variable_table.jsx
```diff
@@ -52,6 +65,7 @@ var VariableTable = React.createClass({
 					<tr>
 						<th className="variable" dangerouslySetInnerHTML={{__html:this.props.variableHeading}} />
 						{valueHeading}
+						<th className="status"></th>
```
mateusz: I'm wondering if it would make sense not to even include the column if no read-only items are provided?

PR #554, opened 2016-04-20
Added a warning about signed commit issues

assertchris:


madmatt:
This will only work if pipelining is enabled. You'll also have to change `DeployForm` to include the message I believe for environments that don't have pipelines enabled or configured.

Note: It's been years since I've touched Deploynaut code, I could well be wrong :)

assertchris:
Could someone else confirm this? I have no idea of how Deploynaut is structured - this was the only template change I could infer from the URL I use to deploy...

@hafriedlander @halkyon @stojg @mateusz

stojg:
Looks like one possible solution for the older version of deploynaut. But I would be keen to get an related issue. I saw @madmatt opened an issue in the gitlib package at https://github.com/gitonomy/gitlib/issues/108

madmatt:
Yeah, I was thinking that gitlib was some kind of PHP extension, but it's actually just a PHP script shelling out to `git`, so it's much easier for us to just fix ourselves :)

halkyon:
This looks like it can be closed, as the issue was fixed upstream in gitonomy. @stojg Can you confirm?

stojg:
I'm closing it, it's fixed as far as I can tell. If not, feel free to reopen.

PR #575, opened 2016-07-22
Abort button

mateusz:


code/control/DNRoot.php
```diff
@@ -1563,6 +1565,32 @@ public function deploylog(SS_HTTPRequest $request) {
...
 	}
 
+	public function abortDeploy(SS_HTTPRequest $request) {
+		$params = $request->params();
+		$deployment = DNDeployment::get()->byId($params['Identifier']);
+
+		if(!$deployment || !$deployment->ID) {
+			throw new SS_HTTPResponse_Exception('Deployment not found', 404);
+		}
+		if(!$deployment->canView()) {
```
mateusz: Needs better security - admin only, or perm code check

sass/style.sass
```diff
@@ -1227,3 +1227,50 @@ textarea.deploy-key
 .content .variables-description p
 	max-width: none
 
+button.abort
```
halkyon: There's a lot of specific styling here, I reckon you should try to remove it and rely on `.btn` and only add what is absolutely necessary (float: right, etc?) here.

mateusz: Done - I still want to tweak the size because.

halkyon: Can it be done with a pre-defined class instead? http://v4-alpha.getbootstrap.com/components/buttons/#sizes

javascript/deploynaut.js
```diff
@@ -275,4 +275,23 @@
 		});
 
 	});
+
+	$('button.abort').click(function(event) {
```
halkyon: `on('click', function(event) {` would be better here, as that's recommended over calling click directly.

code/control/DNRoot.php
mateusz: Done.

javascript/deploynaut.js
mateusz: ok then I'm not doing it :-)

sass/style.sass
mateusz: Nah, I want it margin-less so it fits inline with the label on the left, and really small, so it doesn't collapse the log. I don't want to fix up the entire page :-| WILL_NOT_PERFORM?

mateusz:
screen shot 2016-07-22 at 4 00 47 pm

code/jobs/DeployJob.php
```diff
@@ -10,7 +10,80 @@ class DeployJob extends DeploynautJob {
...
+			])) {
+				echo sprintf(
+					'[-] Signal "%s" received, delivering to own process group, PID "%s".' . PHP_EOL,
+					$signal,
+					getmypid()
+				);
+				// Mark the signal as received.
+				unlink($sigFile);
+				// Dispatch to own process group.
+				$pgid = posix_getpgid(getmypid());
```
stojg: double check that the pgid is not zero, bad things can happen since pid zero is the init system.

mateusz: Fixed

PR #579, opened 2016-07-31
Introduce a state machine to DNDeployment, and rewrite state transitions

mateusz:


code/control/DNDeploymentDispatcher.php
```diff
@@ -0,0 +1,96 @@
...
+			return $this->project404Response();
+		}
+
+		$env = $this->getCurrentEnvironment($project);
+		if(!$env) {
+			return $this->environment404Response();
+		}
+
+		$machine = $this->getMachine();
+		$machine->apply($request->param('State'));
```
stojg: what will happen (due to cockups) if we actually can't transition because the `State` param is wrong or null and so on?

code/control/DNDeploymentDispatcher.php
```diff
@@ -0,0 +1,96 @@
...
+				case 'approve':
+					$code = self::ALLOW_DNDEPLOYMENT_APPROVE;
+					break;
+				case 'queue':
+					$code = self::ALLOW_DNDEPLOYMENT_QUEUE;
+					break;
+				}
+
+				if (!$code || !$project->allowed($code)) {
+					$e->reject();
```
stojg: throws exception?

code/model/jobs/DNDeployment.php
```diff
@@ -52,21 +52,73 @@ public function getTitle() {
...
-			false => "Invalid",
-		);
-		return $remap[$int];
+	public function getFiniteState() {
+        return $this->State;
+    }
+
+    public function setFiniteState($state) {
+        $this->State = $state;
+		$this->write();
```
stojg: whitespace

code/model/jobs/DNDeployment.php
```diff
@@ -52,21 +52,73 @@ public function getTitle() {
...
-	public static function map_resque_status($int) {
-		$remap = array(
-			Resque_Job_Status::STATUS_WAITING => "Queued",
-			Resque_Job_Status::STATUS_RUNNING => "Running",
-			Resque_Job_Status::STATUS_FAILED => "Failed",
-			Resque_Job_Status::STATUS_COMPLETE => "Complete",
-			false => "Invalid",
-		);
-		return $remap[$int];
+	public function getFiniteState() {
```
stojg: whitespace

code/model/jobs/DNDeployment.php
```diff
@@ -52,21 +52,73 @@ public function getTitle() {
...
+        return $this->State;
+    }
+
+    public function setFiniteState($state) {
+        $this->State = $state;
+		$this->write();
+    }
+
+	public function getStatus() {
```
stojg: whitespace

code/jobs/DeployJob.php
```diff
@@ -132,23 +132,27 @@ public function perform() {
...
+	public function onFailure(Exception $exception) {
+		$this->updateStatus('fail');
+	}
+
 	/**
-	 * @param string $status
+	 * @param string $status Transition
 	 * @global array $databaseConfig
 	 */
 	protected function updateStatus($status) {
```
stojg: Should this be renamed to `transitionTo` to make it clearer that it triggers a state change (if possible), or is that to disruptive with all the other jobs?

code/control/DNDeploymentDispatcher.php
```diff
@@ -0,0 +1,96 @@
...
+	private static $url_handlers = [
+		'$Id/apply/$State' => 'apply',
+		'$Id/can/$State' => 'canApply',
+	];
+
+	public function getModel($name) {
+		$id = $this->request->param('Id');
+		$deployment = DNDeployment::get()->byId($id);
+		return [
+			'Status' => $deployment->Status
```
stojg: $deployment->State? the $deployment->getStatus() method is just there for legacy compat?

code/control/DNDeploymentDispatcher.php
```diff
@@ -0,0 +1,96 @@
...
+		if(!$env) {
+			return $this->environment404Response();
+		}
+
+		return $this->asJSON([
+			'Can' => $this->getMachine()->can($request->param('State'))
+		]);
+	}
+
+	private function getMachine() {
```
stojg: visibility private?

code/control/DNDeploymentDispatcher.php
```diff
@@ -0,0 +1,96 @@
...
+		$id = $this->request->param('Id');
+		$deployment = DNDeployment::get()->byId($id);
+		$project = $deployment->Environment()->Project();
+		$machine = $deployment->getMachine();
+
+		// Map authorisation rules to selected transitions.
+		$machine->getDispatcher()->addListener(
+			'finite.test_transition',
```
stojg: ?

code/control/DNDeploymentDispatcher.php
mateusz: It will blow up in your face with exception ;-) I decided not to spend too much time here, because I don't have any actually use cases for this dispatcher yet. The whole API might be thrown away after deliberation.

code/control/DNDeploymentDispatcher.php
mateusz: Let me add a comment here.

code/control/DNDeploymentDispatcher.php
mateusz: Good catch.

code/control/DNDeploymentDispatcher.php
mateusz: Making it protected then.

code/jobs/DeployJob.php
mateusz: The latter - it's an API enforced by the DeploynautJob. And the other jobs are still using "Status" instead of state machines. Unfortunate, but forgivable I think

code/model/jobs/DNDeployment.php
mateusz: grumble grumble PSR

code/control/DNDeploymentDispatcher.php
mateusz: not from here, that's just a flag setter. The state machine implementation throws it later though.

PR #580, opened 2016-08-01
Fixing border cases where deployment may not exist. Show 404.

halkyon:


assertchris:
Not so much fix as avoid...

PR #587, opened 2016-08-02
Autumn cleaning on frontend

stojg:


js/form.jsx
```diff
@@ -114,6 +117,9 @@ var Form = React.createClass({
 			return true;
 		}
 
+		//if (!component.isMounted()) {
```
halkyon: Delete this?

PR #602, opened 2016-08-15
Allow backup database before deployment. Open advanced options up.

halkyon:


code/control/DNRoot.php
```diff
@@ -1096,14 +1096,6 @@ public function gitRevisions(SS_HTTPRequest $request) {
 			return $this->environment404Response();
 		}
 
-		// For now only permit advanced options on one environment type, because we hacked the "full-deploy"
```
stojg: so.. this is going to give everybody the option of 'full deploy and no rollback'?

halkyon: yeah, I think I need to add something so each environment has a `supported_advanced_features` or something. Any suggestions?

code/jobs/DeployJob.php
```diff
@@ -124,8 +145,14 @@ public function onFailure(Exception $exception) {
...
 		global $databaseConfig;
 		DB::connect($databaseConfig);
-		$dnDeployment = DNDeployment::get()->byID($this->args['deploymentID']);
-		$dnDeployment->getMachine()->apply($status);
+		$deployment = DNDeployment::get()->byID($this->args['deploymentID']);
+		$deployment->getMachine()->apply($status);
+
+		$backup = $deployment->BackupDataTransfer();
+		if($backup && $backup->exists()) {
+			$backup->Status = $status;
```
mateusz: Watch out, `Status` is different to `State`. This function gets passed the `State`, but is named to fit with the old API. The "$status" should probably be renamed to "$state". You might need to map from `DNDeployment::TR_COMPLETE` to Backup::Status

halkyon: true, I did not see that one coming. Damn :/

code/jobs/DeployJob.php
```diff
@@ -80,6 +80,22 @@ public function perform() {
 		$deployment = DNDeployment::get()->byID($this->args['deploymentID']);
 		$environment = $deployment->Environment();
 		$project = $environment->Project();
+		$backupDataTransfer = null;
+		$backupMode = !empty($this->args['backup_mode']) ? $this->args['backup_mode'] : 'db';
+
+		if(!empty($this->args['backup_before_deploy'])) {
```
mateusz: I have a suggestion: how about this option is controlled on three levels:
1) backend-specific default ("on" for Capistrano, "off" for Rainforest)
2) overriden by CMS per-environment setting 
3) overriden by user in "advanced options"

We can then skip the implementation of (3) because the UI is changing anyway and maybe people don't even need it - we could try shipping it and adding it later if someone complains :-)

code/jobs/DeployJob.php
```diff
@@ -80,6 +80,22 @@ public function perform() {
...
+			$backupDataTransfer = DNDataTransfer::create();
+			$backupDataTransfer->EnvironmentID = $environment->ID;
+			$backupDataTransfer->Direction = 'get';
+			$backupDataTransfer->Mode = $backupMode;
+			$backupDataTransfer->ResqueToken = $deployment->ResqueToken;
+			$backupDataTransfer->AuthorID = $deployment->DeployerID;
+			$backupDataTransfer->write();
+
+			$deployment->BackupDataTransferID = $backupDataTransfer->ID;
```
mateusz: How about we comment the reason why we trigger a job from a job so we don't have to reverse engineer the reasoning next time around? Might make sense to write an essay next to the $db definition of the BackupDataTransferID.

code/control/DNRoot.php
halkyon: Updated

code/jobs/DeployJob.php
halkyon: Added some text. Want to have another look?

code/jobs/DeployJob.php
```diff
@@ -96,6 +116,17 @@ public function perform() {
 				throw new \RuntimeException($message);
 			}
 
+			if($backupDataTransfer) {
+				$log->write('Backing up existing data');
+				try {
+					$environment->Backend()->dataTransfer($backupDataTransfer, $log);
+					$this->updateBackupStatus('Finished');
```
mateusz: I'd probably set the status directly on $backupDataTransfer->Status directly here.

code/jobs/DeployJob.php
```diff
@@ -96,6 +116,17 @@ public function perform() {
 				throw new \RuntimeException($message);
 			}
 
+			if($backupDataTransfer) {
```
mateusz: Maybe pull the whole block into a separate func at least

PR #604, opened 2016-08-17
Deploy ui summary

stojg:
# Add the first version of the deployment diff summary

When users pick a git sha it asks the backend for the deploy plan / diff
and will display that on step 2 of the deployment form.

This change also includes a very simple complete and loading hints on
the step navigation.

# Extract reducers into their own modules

The root reducer were becoming to big to easily understand and change. This change replaces one
reducer with the redux combineReducers() that only permits reducers to manage a certain slice of
the global state.


code/control/DeployPlanDispatcher.php
```diff
@@ -178,6 +179,76 @@ protected function createFetch() {
...
+		}
+
+		// Performs canView permission check by limiting visible projects
+		$environment = $this->getCurrentEnvironment($project);
+		if(!$environment) {
+			return $this->environment404Response();
+		}
+
+		if(!trim($request->getBody())) {
+			$output = [
```
halkyon: Pretty minor, but you could save a few lines of code by just passing the output array straight into getAPIResponse rather than make a variable out of it, you don't use it anywhere else :)

js/components/Messages.jsx
```diff
@@ -0,0 +1,35 @@
...
+	var messages = props.messages.map(function(msg, index) {
+		return <Message key={index} message={msg.text} type={msg.code} />;
+	});
+	return (
+		<div>
+			{messages}
+		</div>
+	);
+}
+
```
halkyon: Extra newline here isn't needed

js/components/DeployDiff.jsx
```diff
@@ -0,0 +1,24 @@
...
+			</div>
+		);
+	}
+
+	return (
+		<SummaryTable {...props} />
+	);
+};
+
+DeployDiff.propTypes = {
```
halkyon: No propTypes here? Is this still needed?

js/_api.js
```diff
@@ -14,13 +14,24 @@ function sleep(time) {
...
-		Accept: 'application/json',
-		'Content-Type': 'application/json'
-	})
+		headers: {
+			Accept: 'application/json',
+			'Content-Type': 'application/json'
+		}
+	};
+
+	if(payload) {
```
halkyon: In what case would payload be empty?

halkyon: Nevermind, seems there are some cases of using `call` without it

code/control/DeployPlanDispatcher.php
```diff
@@ -178,6 +179,49 @@ protected function createFetch() {
 	}
 
 	/**
+	 * @param SS_HTTPRequest $request
+	 *
+	 * @return SS_HTTPResponse
+	 */
+	public function deploysummary(SS_HTTPRequest $request) {
+
```
halkyon: I'd suggest not adding newlines on the first line of functions if you can help it, it's kind of annoying (IMHO)

js/_api.js
```diff
@@ -14,13 +14,24 @@ function sleep(time) {
...
+		var data = new FormData();
+		data.append("json", JSON.stringify(payload));
+		options.body = JSON.stringify(payload);
+	}
+
+	return fetch(`${uri}`, options)
```
halkyon: Don't think you need the backticks no? Since it's just using $uri which is a string directly.

js/containers/App.jsx
```diff
@@ -120,11 +108,22 @@ function App(props) {
...
+		isLoading: [
+			state.git.is_loading || state.git.is_updating,
+			state.plan.is_loading,
+			false,
+			false
+		],
+		isFinished: [
+			state.git.selected_ref !== "",
+			state.plan.validation_code === 'success',
+			state.plan.validation_code === 'success' && (state.approval.approved || state.approval.bypassed),
```
halkyon: What is the bypassed state?

code/control/DeployPlanDispatcher.php
```diff
@@ -178,6 +179,49 @@ protected function createFetch() {
...
+		}
+
+		$options = [
+			'sha' => $jsonBody['sha']
+		];
+
+		$strategy = $this->environment->Backend()->planDeploy($this->environment, $options);
+		$data = $strategy->toArray();
+
+		$interface = $this->project->getRepositoryInterface();
```
halkyon: Oh dear. This thing again... I wonder if we could encapsulate that into `RepositoryInterface`, or put it somewhere else...

js/containers/App.jsx
```diff
@@ -120,11 +108,22 @@ function App(props) {
 
 const mapStateToProps = function(state) {
 	return {
-		message: state.message,
-		messageType: state.message_type,
-		activeStep: state.activeStep,
+		isLoading: [
+			state.git.is_loading || state.git.is_updating,
+			state.plan.is_loading,
+			false,
```
halkyon: What are these false arguments?

code/control/DeployPlanDispatcher.php
stojg: I thought about it, but `RepositoryInterface` is just an `ArrayData` generated by DNProject and I didn't wanna go too far with this copy pasta

js/containers/App.jsx
stojg: skip the whole approval process, go straight to start, collect $200 and a deploy

js/containers/App.jsx
stojg: placeholders, they will never show a 'loading' spinner at the moment.

stojg: tbh, this whole section is a bit of a place holder until we know how it will connect with the 'real' menu

js/components/DeployDiff.jsx
stojg: Added a proptype

js/components/Messages.jsx
stojg: gone

PR #609, opened 2016-08-24
NEW Make deploynaut admin the default admin tab

dhensby:
This makes it so the deploynaut admin is the default section of the CMS to load when going to `/admin` in your browser

### DIFF
--- a/_config/config.yml
+++ b/_config/config.yml
@@ -1,6 +1,8 @@
 ---
 Name: deploynautconfig
 ---
+AdminRootController:
+  default_panel: 'DNAdmin'
 DNProject:
   defaults:
     DiskQuotaMB: "1024"


andrewandante:
Woot! Tested locally and it works, only question is whether this should be here or defined in a higher-level config.

dhensby:
> whether this should be here or defined in a higher-level config.

Basically, should a project make the decision about which admin is default or  should this module impose it on the projects

halkyon:
Thanks @dhensby. @andrewandante I think it's a good default, most of the time going into admin when using deploynaut you'll want to be managing projects and environments, and that config can be overridden in more specific config anyway.

halkyon:
By the way, deploynaut imposes basically everything on a project as it stands, it overrides all the routing so that "normal" page requests don't work anyway. It's not really a module more than it is a project in itself :)

PR #610, opened 2016-08-28
Default predeploy backup option to production envs only

halkyon:


code/backends/CapistranoDeploymentBackend.php
```diff
@@ -154,9 +154,9 @@ public function deploy(
 	/**
 	 * @return array
 	 */
-	public function getDeployOptions() {
+	public function getDeployOptions(DNEnvironment $environment) {
 		return [
-			new PredeployBackupOption(true),
+			new PredeployBackupOption($environment->Usage === 'Production'),
```
stojg: We should change the magic string to a constant for `Usage`, before we start mispeling `prod`, `Production` or `production`

code/backends/DemoDeploymentBackend.php
```diff
@@ -67,7 +67,7 @@ public function deploy(
 	/**
 	 * @return array
 	 */
-	public function getDeployOptions() {
+	public function getDeployOptions(DNEnvironment $environment) {
```
stojg: missing docblock for argument

code/backends/CapistranoDeploymentBackend.php
halkyon: Done!

code/backends/DemoDeploymentBackend.php
halkyon: Done!

PR #613, opened 2016-08-30
add CSRF checking on plan deploy dispatcher

stojg:


code/control/DeployPlanDispatcher.php
```diff
@@ -74,6 +74,7 @@ public function index(\SS_HTTPRequest $request) {
 	public function gitupdate(SS_HTTPRequest $request) {
 		switch($request->httpMethod()) {
 			case 'POST':
+				$this->checkSecurityToken(sprintf('%sSecurityID', get_class($this)));
```
stojg: I'm not super happy with the result of a failing check is a text response, I was thinking of making changes to it so it would return a response that is more suitable for js consumption to display to the user.

js/_api.js
```diff
@@ -20,21 +20,27 @@ export function call(uri, method, payload) {
...
-			'Content-Type': 'application/json'
+			Accept: 'application/json'
 		}
 	};
 
-	if(payload) {
-		var data = new FormData();
-		data.append("json", JSON.stringify(payload));
-		options.body = JSON.stringify(payload);
+	if(method.toLowerCase() === "post") {
```
stojg: I regard all post calls to send the csrf token back, since they should by some standards of http APIs create/update/delete resources. I currently don't see any need to send csrf tokens on GET calls.

code/control/DeployPlanDispatcher.php
mateusz: I believe you'll need to reset the security token from time to time to achieve Premium Security (`$this->getSecurityToken()->reset();`). The way this is implemented, if you never call `asJSON`, the token will remain the same for the entire user session.

code/control/DeployPlanDispatcher.php
```diff
@@ -290,6 +284,10 @@ protected function getGitPrevDeploys($project) {
 	 * @return SS_HTTPResponse
 	 */
 	protected function getAPIResponse($output, $statusCode) {
```
mateusz: Do you promise to push this functionality back into a dispatcher?

js/Plan.jsx
```diff
@@ -22,7 +20,8 @@ var store = Redux.createStore(
 
 var Plan = function(props) {
 
-	window.api_url = props.model.APIEndpoint;
+	window.api_url = props.model.api_endpoint;
```
mateusz: Will we ever use more than one Dispatcher?

js/_api.js
```diff
@@ -20,21 +20,27 @@ export function call(uri, method, payload) {
...
+		options.body = new FormData();
+		options.body.append(window.api_auth.name, window.api_auth.value);
+
+		var data = payload || {};
+		Object.keys(data).forEach(function(key) {
+			options.body.append(key, payload[key]);
+		});
 	}
 
 	return fetch(uri, options)
```
mateusz: Now that I look at it, how will module writers be able to add their own API definitions? Maybe the `getRevisions`, `updateRepo` and so on don't belong here really?

js/_api.js
```diff
@@ -20,21 +20,27 @@ export function call(uri, method, payload) {
...
+			Accept: 'application/json'
 		}
 	};
 
-	if(payload) {
-		var data = new FormData();
-		data.append("json", JSON.stringify(payload));
-		options.body = JSON.stringify(payload);
+	if(method.toLowerCase() === "post") {
+		options.body = new FormData();
```
mateusz: Didn't we have a massive war about how we submit data to the backend? Didn't JSON win over form-encoding by any chance? :-)

code/control/DeployPlanDispatcher.php
stojg: ah, good catch, I had it in there, but in my prep for making it a PR it vanished.

code/control/DeployPlanDispatcher.php
stojg: Probably will have to, since I think my next step is to break this up into dispatchers for 
- git stuff
- approval
- deploy planning
- deploy

js/Plan.jsx
stojg: Yes, I was thinking that this enpoint would probably end up as 
`http://deploy.silverstripe.dev/naut/project/cbmicro/environment/Production/`
and the javascript api module would then add the dispatcher name to the url for the individual calls.

current api_endpoint is locked down to `http://deploy.silverstripe.dev/naut/project/cbmicro/environment/Production/plan`

js/_api.js
stojg: Massive war? I discovered that the current functionality for token checking is using the built in silverstripe `$request->postVar('asdsd');` and has no ability to easily read json body content from the content. As my trusty peer reviewer and friend, would you prefer?

js/_api.js
stojg: Yeah, I'm a little bit uncertain about where the API should be defined. Or actually, I'm pretty much clueless. :/

js/_api.js
stojg: One alternative is to standardise on json and add this to the The Dispatcher

```
/**
 * @return array
 */
function getRequestBody($request) {
    if(!trim($request->getBody())) {
         return $this->getAPIResponse(['message' => 'no body was sent in the request'], 400);
    }   
    $jsonBody = json_decode($request->getBody(), true);
    if(empty($jsonBody)) {
        return $this->getAPIResponse(['message' => 'request did not contain a parsable JSON payload'], 400)   
    }
    return $jsonBody;
}
```

And maybe make https://github.com/silverstripe/deploynaut/blob/master/code/control/Dispatcher.php#L51 check both `postVar` and the json body as a backward compatible way?

code/control/DeployPlanDispatcher.php
mateusz: 👍

code/control/DeployPlanDispatcher.php
mateusz: Yeah, that's why I'm asking.

js/Plan.jsx
mateusz: oh sorry, the comment was to the line below: if you store the api_auth in window, what will happen if you have multiple dispatchers? Seems to me like the token state will go out of sync, especially that we are using the dispatcher ClassName in the token name?

js/_api.js
mateusz: That was implicitly understood 👍

code/control/DeployPlanDispatcher.php
```diff
@@ -290,6 +284,12 @@ protected function getGitPrevDeploys($project) {
...
+		$secToken = $this->getSecurityToken();
+		$secToken->reset();
```
mateusz: Follow up question about resetting then: if we are not submitting the token as part of GET, should we then not reset the token on GET handlers in the backend? I'm slightly worried about concurrency: could this blow up if a GET request is followed by a POST in quick succession?

js/_api.js
mateusz: You are making history. Well, I'd say at minimum the api helpers shouldn't be colocated with the API calls itself. We can standardise on an _api.js per module though for defining the actual permitted calls?

js/_api.js
mateusz: Hrm not making myself clear - the old Dispatcher packed everything into JSON "Details", and then unwrapped them from these Details. I guess that's not really important. I'm happy with whatever you recommend - I just remembered with had a bit of a discussion with Michael whether we submit everything via JSON, or we submit form-encoded, and your opinion was on the side of use-JSON-always :-)

code/control/DeployPlanDispatcher.php
stojg: that is an excellent point that I should have thought off. :headdesk: will find a better solution.

code/control/DeployPlanDispatcher.php
```diff
@@ -124,15 +125,18 @@ public function gitrefs(\SS_HTTPRequest $request) {
...
-	 * @param string $name of the component
+	 * @param string $name
 	 *
 	 * @return array
 	 */
-	public function getModel($name) {
+	public function getModel($name = '') {
 		return [
-			'APIEndpoint' => Director::absoluteBaseURL().$this->Link()
+			'namespace' => self::ACTION_PLAN,
```
stojg: Send over information to the front end about which dispatcher this is, so we can have multiple dispatchers and now mess up the CSRF tokens

js/Plan.jsx
```diff
@@ -21,8 +20,11 @@ var store = Redux.createStore(
 );
 
 var Plan = function(props) {
-
-	window.api_url = props.model.APIEndpoint;
+	store.dispatch(api.setAPI(
+		props.model.namespace,
+		props.model.api_endpoint,
+		props.model.api_auth
+	));
```
stojg: Push (hydrate) the store with initial api/dispatcher info

js/_actions.js
```diff
@@ -54,9 +58,9 @@ export function failRevisionsGet(err) {
 }
 
 export function getRevisions() {
-	return (dispatch) => {
+	return (dispatch, getState) => {
 		dispatch(startRevisionGet());
-		return api.getRevisions()
+		return api.call(dispatch, getState, '/gitrefs', 'get')
 			.then(json => dispatch(succeedRevisionsGet(json)))
```
stojg: The action themselves know which dispatcher they are connected to, so they should now which url they would like to call.

js/_api.js
```diff
@@ -10,31 +11,74 @@ export function APIError(message, json) {
...
 	return new Promise((resolve) => setTimeout(resolve, time));
 }
 
-export function call(uri, method, payload) {
+var namespace = '';
+export function setNamespace(ns) {
+	namespace = ns;
+}
+
+export function call(dispatch, getState, uri, method, payload) {
```
stojg: Since the CSRF token, api endpoint and other dispatcher specific information is in the store, we need to dispatch new information when it's recieved and also being able to get the state so we can include the CSRF in the request.

code/control/DeployPlanDispatcher.php
stojg: I thought about this.  

1) I remove the regeneration of the CSRF on responses to GET requests
2) we can only make one POST request at the same time to the same dispatcher.

Now I'm not really sure if I want to solve the 2) in this PR, or if it's going to be a common problem. Hopefully we can break up potential POST actions into their own dispatchers.

stojg:
@mateusz the mystery of the namespaced security tokens has been found. I asked the author and it was to make sure that the react front-end don't interfere with any static default SilverStripe forms. I see three options:

1) revert that and use the default token name, we cannot show static forms and dispatcher forms on the same url. Works now, but will a huge WTF if that happens in the future.
2) Hard code the dispatcher security token name do be different from the default. We would need a full regression testing I would assume to ensure that we don't break current react components
3) namespace it like the above.

I'm leaning against option 2 or 3.

code/control/DeployPlanDispatcher.php
```diff
@@ -290,6 +285,16 @@ protected function getGitPrevDeploys($project) {
 	 * @return SS_HTTPResponse
 	 */
 	protected function getAPIResponse($output, $statusCode) {
+		// GET and HEAD requests should not change state and therefore
```
stojg: @mateusz after our discussion about the racy condition of GET and POST at the same time, I've added a check below to ensure only non GET and HEAD requests resets the token.

mateusz: 👍

mateusz:
Hrm so we still have a problem of two tabs using the same dispatcher ClassName.

Looking at OWASP, it seems though there are acceptable ways of mitigating CRSF without having to maintain state on the server, like the [Double Submit Cookie](https://www.owasp.org/index.php/Cross-Site_Request_Forgery_(CSRF)_Prevention_Cheat_Sheet#Synchronizer_.28CSRF.29_Tokens). It solves all state desynchronisation problems once and for all. Could we ditch the whole CRSF management malarky (back and frontend) and use something like [this](https://github.com/mebjas/CSRF-Protector-PHP)?

Other than that I'm starting to warm up to option (3), make all Dispatchers use the same one global DispatcherToken that is never changed unless you log in or out. Maintaining parallel dispatcher state feels like asking for trouble :-(

Sorry for back and forth on that. @halkyon care to give a second opinion?

PR #614, opened 2016-09-01
Add new deployment button, add modal open action

halkyon:


js/containers/App.jsx
```diff
@@ -85,23 +86,27 @@ function App(props) {
...
-						value={props.activeStep}
-						onClick={props.onTabClick}
-					/>
-				</div>
-				<div className="col-md-9">
-					{content}
+		<div>
+			<NewDeployPlan />
+
+			<Modal show={props.isOpen} closeHandler="" title="Deployment">
```
stojg: closeHandler? that should be set in the `mapDispatchToProps` so if the modal closes itself it can set the navigation.state = false.

js/Modal.jsx
```diff
@@ -3,19 +3,25 @@ var ReactDOM = require("react-dom");
...
-	getInitialState: function() {
+	getDefaultProps: function() {
 		return {
-		};
+			show: true
+		}
+	},
+
+	componentWillReceiveProps: function(props) {
+		this.modal(props);
```
stojg: I don't think the "calling" code should know about what options goes into a bootstrap modal. In this case it's basically the same, but I prefer to map it explicitly.
```
this.modal({show: props.show});
```

btw, what happened with keyboard: true?

halkyon:
@stojg comments fixed, thanks for checking it out.

js/Modal.jsx
```diff
@@ -3,19 +3,25 @@ var ReactDOM = require("react-dom");
...
+		this.modal(this.props);
 	},
 
 	componentWillUnmount: function() {
 		this.modal('hide');
 	},
 
+
 	modal: function(options) {
 		$(ReactDOM.findDOMNode(this)).modal(options);
```
stojg: the ref syntax has changed a bit since when this component was written, and findDOMNode is no longer prefered way todo this. https://facebook.github.io/react/docs/top-level-api.html#reactdom.finddomnode

There is a reference in the DropDown.jsx that solves 

First create selector var https://github.com/silverstripe/deploynaut/blob/master/js/components/DropDown.jsx#L47

In the render of Modal, set a ref argument that looks something like this

https://github.com/silverstripe/deploynaut/blob/master/js/components/DropDown.jsx#L92

Now all places in this file that used to do `ReactDOM.findDOMNode(this)` can be replaced with `this.selector`, like this https://github.com/silverstripe/deploynaut/blob/master/js/components/DropDown.jsx#L62

PR #620, opened 2016-09-02
Fix password resets email

stojg:
For some reason when the code asks for a Member_ForgotPasswordEmail::create() the injector always return the base class `Email`


stojg:
This has been broken since https://github.com/silverstripe/deploynaut/commit/22901a8ce87001f7162c9e757f9724bede09283d

PR #622, opened 2016-09-04
[wip] Adding deploy history to new deploy plan page

halkyon:


js/_actions.js
```diff
@@ -63,6 +63,22 @@ export function failRevisionsGet(err) {
 	};
 }
 
+export const SUCCEED_DEPLOY_HISTORY_GET = 'SUCCEED_DEPLOY_HISTORY_GET';
```
stojg: You might want to get a STARTED_* action, so you can display a loader indicator. For example during initial load and for pagination.

js/containers/DeployHistory.jsx
```diff
@@ -0,0 +1,51 @@
+var React = require('react');
+var ReactRedux = require('react-redux');
+var actions = require('../_actions.js');
+
+var DeployHistory = function(props) {
+	var tableRows = [];
+
+	for (var i in props.data) {
```
stojg: you might want to
1) add this line below this 
```
     // skip loop if the property is from prototype
     if(!props.data.hasOwnProperty(i)) continue;
```
2) Use Object.keys() and map
```
const tableRows = Object.keys(props.data).map(function(index) {
   return (... ):
});
```

js/containers/DeployHistory.jsx
```diff
@@ -0,0 +1,51 @@
...
+};
+
+const mapStateToProps = function(state) {
+	return {
+		data: state.deployhistory.data
+	};
+};
+
+const mapDispatchToProps = function(dispatch) {
+	dispatch(actions.getDeployHistory());
```
stojg: this should not be here, this section is used for sending props that are callbacks for dispatching actions. For example a refresh button / link.

See Plan.jsx `store.dispatch(actions.getRevisionsIfNeeded());`

js/reducers/deployhistory.js
```diff
@@ -0,0 +1,20 @@
+var _ = require('underscore');
+
+var actions = require('../_actions.js');
+
+module.exports = function deployhistory(state, action) {
+	if(typeof state === 'undefined') {
+		return {
+			data: []
```
stojg: you probably want a `is_loading` state here

code/control/DeployPlanDispatcher.php
```diff
@@ -176,6 +177,25 @@ protected function createFetch() {
...
+				'CreatedDate' => $deployment->Created,
+				'Branch' => $deployment->Branch,
+				'Tags' => $deployment->getTags()->toArray(),
+				'Changes' => $deployment->getDeploymentStrategy()->getChanges(),
+				'CommitMessage' => $deployment->getCommitMessage(),
+				'Deployer' => $deployment->Deployer()->getName(),
+				'State' => $deployment->State,
+			];
+		}
+		return $this->getAPIResponse($data, 200);
```
stojg: replace $data with `['history' => $data]`

PR #625, opened 2016-09-06
Adding pagination to deploy history

halkyon:


code/control/DeployDispatcher.php
```diff
@@ -57,7 +57,16 @@ public function index(\SS_HTTPRequest $request) {
 	 */
 	public function history(SS_HTTPRequest $request) {
 		$data = [];
-		foreach($this->DeployHistory() as $deployment) {
+		$list = $this->DeployHistory();
+		$page = $request->getVar('page') ?: 1;
+		$start = ($page-1) * $list->getPageLength();
```
stojg: Can you do some sanity checking, like is_numeric() and clamp it between [0, -> numRecords]?

js/_actions.js
```diff
@@ -99,10 +99,14 @@ export function getRevisions() {
 	};
 }
 
-export function getDeployHistory() {
+export function getDeployHistory(page) {
+	if (page === undefined) {
```
stojg: don't we need `typeof page` ?

PR #630, opened 2016-09-07
Adding current build status

halkyon:


code/control/DeployDispatcher.php
```diff
@@ -76,31 +77,54 @@ public function history(SS_HTTPRequest $request) {
...
 		if (empty($list)) {
-			return $this->getAPIResponse(['message' => 'No deploy history'], 404);
+			return $this->getAPIResponse(['list' => []], 200);
 		}
 
+		$currentBuild = $this->environment->CurrentBuild();
+
 		foreach ($list as $deployment) {
 			$data[] = [
+				'DeploymentID' => $deployment->ID,
```
stojg: I just noticed this, probably should have noticed it yesterday.. but DeplyomentID vs deplyomentID vs deployment_id naming standards..? which one are we running with?

halkyon: How about just ID? Notice that's how it output in another dispatcher action.

halkyon: Changed to ID for now.

js/containers/CurrentBuildStatus.jsx
```diff
@@ -0,0 +1,37 @@
...
+var actions = require('../_actions.js');
+
+var CurrentBuildStatus = function(props) {
+	var branch = props.build.Branch;
+	var sha = props.build.SHA;
+	var shortSha = "";
+	if (typeof sha === 'string') {
+		shortSha = sha.substring(0,7);
+	}
+	if (typeof branch === 'string') {
```
stojg: Could this instead be done in CSS later since it's only a visual thing?

js/reducers/currentbuild.js
```diff
@@ -0,0 +1,26 @@
...
+	switch (action.type) {
+		case actions.START_CURRENT_BUILD_STATUS_GET:
+			return _.assign({}, state, {
+				is_loading: true
+			});
+		case actions.SUCCEED_CURRENT_BUILD_STATUS_GET:
+			return _.assign({}, state, {
+				build: action.data.build,
+				is_loading: false
+			});
```
stojg: You will add a FAIL_CURRENT_BUILD_STATUS_GET later?

js/containers/CurrentBuildStatus.jsx
halkyon: Yup. Updated.

js/reducers/currentbuild.js
halkyon: yeah,the deploy history is missing this too. :/

halkyon: There is no error handling for either of these components, so probably good to look at that next. :)

code/control/DeployDispatcher.php
stojg: Sorry, I wasn't clear. I was talking about all of the data and the format of them that we send to the front end. What is our standard?

code/control/DeployDispatcher.php
```diff
@@ -76,31 +77,54 @@ public function history(SS_HTTPRequest $request) {
...
-			return $this->getAPIResponse(['message' => 'No deploy history'], 404);
+			return $this->getAPIResponse(['list' => []], 200);
 		}
 
+		$currentBuild = $this->environment->CurrentBuild();
+
 		foreach ($list as $deployment) {
 			$data[] = [
+				'ID' => $deployment->ID,
 				'CreatedDate' => $deployment->Created,
```
stojg: Sorry, I wasn't clear. I was talking about all of the data and the format of them that we send to the front end. What is our naming convention?

upper camel case
use lower camel case
SCREAMING_SNAKE_CASE?

code/control/DeployDispatcher.php
```diff
@@ -76,31 +77,54 @@ public function history(SS_HTTPRequest $request) {
...
 				'Approver' => $deployment->Approver()->getName(),
 				'State' => $deployment->State,
+				'IsCurrentBuild' => $currentBuild ? ($deployment->ID === $currentBuild->ID) : null
 			];
 		}
 
-		return $this->getAPIResponse(
-			[
-				'list' => $data,
-				'pagelength' => $list->getPageLength(),
```
stojg: here the naming convention is different again

code/control/DeployDispatcher.php
halkyon: I think upper camel case for now, so that it matches the current DNDeployment etc field names in the job models.

code/control/DeployDispatcher.php
halkyon: Questions about naming, they are so hard! ;-) so what is your opinion then?

code/control/DeployDispatcher.php
stojg: Should I change all the other dispatchers? Because I want the API to be consistent?

halkyon: Maybe merge this first then change everything? I can do it if you want

stojg: I don't want to make a huge thing out of it, but I've never seen successful API that changes it output format . 🐹

halkyon: I'm all for consistency, of course.

PR #634, opened 2016-09-08
Open existing deployment details from history or current build

halkyon:


code/control/DeployDispatcher.php
```diff
@@ -80,66 +80,74 @@ public function history(SS_HTTPRequest $request) {
...
+	/**
+	 * @param \SS_HTTPRequest $request
+	 * @return \SS_HTTPResponse
+	 */
+	public function show(SS_HTTPRequest $request) {
+		$deployment = DNDeployment::get()->byId($request->param('ID'));
+		if (!$deployment || !$deployment->exists()) {
+			return $this->getAPIResponse(['message' => 'This deployment does not exist'], 404);
+		}
+		if (!$deployment->canView()) {
```
stojg: I guess this will have to be rethought with approvers and other type of users later?

PR #650, opened 2016-09-13
Don't create a new deployment on deploy

stojg:


code/control/DeployDispatcher.php
```diff
@@ -172,25 +172,31 @@ public function save(\SS_HTTPRequest $request) {
 	 * @return \SS_HTTPResponse
 	 */
 	public function start(SS_HTTPRequest $request) {
+
+		if (strtolower($request->httpMethod()) !== 'post') {
```
halkyon: httpMethod is always uppercase, so why bother lowering it?

code/control/DeployDispatcher.php
```diff
@@ -172,25 +172,31 @@ public function save(\SS_HTTPRequest $request) {
 	 * @return \SS_HTTPResponse
 	 */
 	public function start(SS_HTTPRequest $request) {
+
```
halkyon: personally I don't like the additions of newlines in the first line of methods... :)

code/control/DeployDispatcher.php
stojg: well, look at that, I didn't now SS did that. nifty.

stojg:
Feedback dealt with

PR #653, opened 2016-09-15
Deploy upcoming

halkyon:


code/model/DNEnvironment.php
```diff
@@ -614,13 +614,24 @@ public function CurrentBuild() {
...
-	 * A history of all builds deployed to this environment
-	 *
+	 * A list of past deployments.
 	 * @return ArrayList
 	 */
 	public function DeployHistory() {
 		return $this->Deployments()
-			->where('SHA IS NOT NULL')
+			->where('"SHA" IS NOT NULL')
+			->filter('State', ['Completed', 'Failed', 'Invalid'])
```
stojg: Can you use the state class const defined?

code/model/DNEnvironment.php
```diff
@@ -614,13 +614,24 @@ public function CurrentBuild() {
...
+	}
+
+	/**
+	 * A list of upcoming or current deployments.
+	 * @return ArrayList
+	 */
+	public function UpcomingDeployments() {
+		return $this->Deployments()
+			->where('"SHA" IS NOT NULL')
+			->filter('State', ['New', 'Submitted', 'Deploying', 'Aborting'])
```
stojg: Can you use the state class const defined?

code/model/DNEnvironment.php
halkyon: so you want the states defined as an array somewhere? You really love your constants don't you :D

stojg: I just dislike magic numbers and strings :) I have fat fingers and it's easy to miss uppercase / lowercase as well. :P

code/model/DNEnvironment.php
halkyon: fixed

code/model/DNEnvironment.php
halkyon: fixed

PR #655, opened 2016-09-19
Fixes to git reference selection with promote builds

halkyon:
- Display fields relating to radio option below radio button
 - Only show promote option when deploying to production environments
 - Add build status information when selecting promote option


code/control/GitDispatcher.php
```diff
@@ -81,17 +81,24 @@ public function update(SS_HTTPRequest $request) {
...
+			$targetEnvironment = DNEnvironment::get()->byId((int) $targetEnvironmentId);
+		}
+
 		$refs = [];
 		$prevDeploys = [];
 
 		$uatEnvironment = $this->project->DNEnvironmentList()->filter('Usage', 'UAT')->first();
 		$uatBuild = $uatEnvironment ? $uatEnvironment->CurrentBuild() : null;
-		if ($uatBuild && $uatBuild->exists()) {
+		if ($uatBuild && $uatBuild->exists() && $targetEnvironment && $targetEnvironment->Usage === 'Production') {
```
stojg: We should probably change to use the `DNEnvironment::PRODUCTION` constant? I noticed that we also haven't used that in other places :(

halkyon: Damnit.. yes, you're right :)

js/reducers/environment.js
```diff
@@ -1,10 +1,23 @@
+var _ = require('underscore');
+
+var actions = require('../_actions.js');
+
 module.exports = function environment(state, action) {
 	if (typeof state === 'undefined') {
 		return {
-			name: ""
+			id: "",
```
stojg: is this an int? should the default be null?

js/reducers/git.js
```diff
@@ -80,7 +80,6 @@ module.exports = function git(state, action) {
 			if (action.list.refs.constructor === Array) {
 				listAsArray = _.assign({}, action.list.refs);
 			}
-			console.log(listAsArray);
```
stojg: oops. sorry.

js/reducers/environment.js
halkyon: yeah, I was rolling with what was already there. I'll make them null.

halkyon: fixed

code/control/GitDispatcher.php
halkyon: fixed

js/reducers/git.js
halkyon: no worries, I was clearing out the easy stuff from lint

PR #657, opened 2016-09-21
Routing for deployment modal

halkyon:


js/containers/DeployModal.jsx
```diff
@@ -112,6 +111,16 @@ const mapStateToProps = function(state) {
...
 	}
 
+	let active_step = 0;
+	if (window.location.hash) {
+		active_step = parseInt(window.location.hash.substring(1), 10);
+	}
+
+	let is_open = false;
+	if (ownProps.params) {
+		is_open = ownProps.params.id !== null || typeof (ownProps.params.id) !== 'undefined';
```
stojg: you are doing a three combination comparison if here, just hidden as one `if` and a double `true comparison` assignment. Change to either use only an if or the comparison to make it more readable.
```
if (typeof (ownProps.params.id) !== 'undefined' && ownProps.params.id !== null) {
   is_open = true;
}
```
or 

```
is_open = typeof (ownProps.params.id) !== 'undefined' && ownProps.params.id !== null;
```

js/_actions.js
```diff
@@ -6,6 +6,8 @@ const gitAPI = api.create('git');
 const planAPI = api.create('plan');
 const deployAPI = api.create('deploys');
 
+export const history = null;
```
halkyon: yeah, it's an exported variable that is set by EnvironmentOverview.jsx. I'll add some comments

js/EnvironmentOverview.jsx
```diff
@@ -1,33 +1,47 @@
...
+const store = Redux.createStore(
+	reducers,
 	Redux.applyMiddleware(
 		thunkMiddleware,
 		createLogger()
 	)
 );
 
-var Plan = function(props) {
+export function getDeploymentHook(store) {
```
halkyon: yeah, that's right. Any ideas on the name? It's basically taking the URL param and getting the deployment (or setting a new deployment).

code/control/EnvironmentOverview.php
```diff
@@ -16,6 +16,10 @@ class EnvironmentOverview extends Dispatcher {
 		self::ACTION_OVERVIEW
 	];
 
+	private static $allowed_actions = [
```
halkyon: not wrong, I'll document

js/EnvironmentOverview.jsx
halkyon: `loadDeployment`?

PR #663, opened 2016-09-26
On API request, ensure user is directed to login if they are logged out

halkyon:


js/_api.js
```diff
@@ -96,6 +96,12 @@ export function create(name) {
 
 				// the http response is in the 200 >=  <= 299 range
 				if (response.ok) {
+					// If the response is to login, such as the user session has timed out, then
```
stojg: So sneaky.

PR #664, opened 2016-09-26
add timezone for users and use it in generating user facing timestamps

jakedaleweb:
This adds the ability to make an `SS_Datetime` object that is timezone specific to what is defined in the logged in Member's settings.

The SS_Datetime objects already defined in `$db` across the code base should be redefined as `SS_Datetimezone` as we go along or as is appropriate for the front end users.

### DIFF
--- a/code/SS_Datetimezone.php
+++ b/code/SS_Datetimezone.php
@@ -1,0 +1,22 @@
+<?php
+/**
+ * Class SS_Datetimezone
+ *
+ * Adds customisable timezones to the nice method on {@link SS_Datetime}.
+ */
+class SS_Datetimezone extends SS_Datetime {
+	/**
+	* Returns the date in the raw SQL-format specific to a given timezone passed from the Member class, e.g. “2006-01-18 16:32:04”
+	*/
+	public function Format($format) {
+		if($this->value){
+			$date = new DateTime($this->value);
+			//if the current user has set a timezone that is not the default then use that
+			$member = Member::currentUser();
+			if ($member && $member->exists() && $member->Timezone && $member->Timezone != date_default_timezone_get()) {
+				$date->setTimezone(new DateTimeZone($member->Timezone));    
+			}
+			return $date->Format($format);
+		}
+	}
+}
--- a/code/extensions/TimezoneMemberExtension.php
+++ b/code/extensions/TimezoneMemberExtension.php
@@ -1,0 +1,27 @@
+<?php
+/**
+ * Class TimezoneMemberExtension.
+ *
+ * Adds all available timezones as an optional field to SilverStripe {@link Member}.
+ */
+class TimezoneMemberExtension extends DataExtension {
+	private static $db = [
+		'Timezone' => 'Varchar(255)',
+	];
+
+	public function getTimezones() {
+		$timezones = timezone_identifiers_list();
+		return array_combine($timezones, $timezones);
+	}
+
+	public function updateCMSFields(FieldList $fields) {   
+		$fields->removeFieldFromTab('Root', 'Timezone');
+		$field = DropdownField::create(
+			'Timezone',
+			'Timezone',
+		  $this->getTimezones())->setEmptyString('For NZ, choose Pacific/Auckland');
+		$fields->addFieldToTab('Root.timezone', $field);
+		return $fields;
+	}
+
+}
--- a/code/model/DNCommit.php
+++ b/code/model/DNCommit.php
@@ -196,12 +196,9 @@ class DNCommit extends ViewableData {
 	public function Created() {
 		$created = $this->commit->getCommitterDate();
 
-		// gitonomy sets the time to UTC, so now we set the timezone to
-		// whatever PHP is set to (date.timezone). This will change in the future if each
-		// deploynaut user has their own timezone
 		$created->setTimezone(new DateTimeZone(date_default_timezone_get()));
 
-		$d = new SS_Datetime();
+		$d = SS_Datetime::create();
 		$d->setValue($created->format('Y-m-d H:i:s'));
 
 		return $d;
--- a/code/model/DNTag.php
+++ b/code/model/DNTag.php
@@ -53,13 +53,11 @@ class DNTag extends ViewableData {
 	public function Created() {
 		$created = $this->tag->getCommit()->getCommitterDate();
 
-		// gitonomy sets the time to UTC, so now we set the timezone to
-		// whatever PHP is set to (date.timezone). This will change in the future if each
-		// deploynaut user has their own timezone
 		$created->setTimezone(new DateTimeZone(date_default_timezone_get()));
 
-		$d = new SS_Datetime();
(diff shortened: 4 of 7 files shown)


halkyon:
@jakedaleweb can you please ensure tabs? The current deploynaut module is all in tabs, so introducing spaces would make things annoying for others :)

code/model/DNTag.php
```diff
@@ -53,13 +53,11 @@ public function SHA() {
 	public function Created() {
 		$created = $this->tag->getCommit()->getCommitterDate();
 
-		// gitonomy sets the time to UTC, so now we set the timezone to
-		// whatever PHP is set to (date.timezone). This will change in the future if each
-		// deploynaut user has their own timezone
 		$created->setTimezone(new DateTimeZone(date_default_timezone_get()));
-
+
```
halkyon: remove this addition of indentation plz :)

code/model/DNCommit.php
```diff
@@ -196,11 +196,8 @@ public function EverDeployedTo($environmentName) {
 	public function Created() {
 		$created = $this->commit->getCommitterDate();
 
-		// gitonomy sets the time to UTC, so now we set the timezone to
-		// whatever PHP is set to (date.timezone). This will change in the future if each
-		// deploynaut user has their own timezone
 		$created->setTimezone(new DateTimeZone(date_default_timezone_get()));
-
+
```
halkyon: remove this addition of indentation plz :)

code/extensions/TimezoneMemberExtension.php
```diff
@@ -0,0 +1,29 @@
+<?php
+/**
+ * Class TimezoneMemberExtension.
+ *
+ * Adds all available timezones as an optional field to SilverStripe {@link Member}.
+ */
+class TimezoneMemberExtension extends DataExtension
+{
```
halkyon: braces should be on the same line as the `class` declaration (at least until we change the formatting of this module to PSR-2)

code/extensions/TimezoneMemberExtension.php
```diff
@@ -0,0 +1,29 @@
...
+        'Timezone' => 'Varchar(255)',
+    ];
+
+    public function getTimezones() {
+        $timezones = timezone_identifiers_list();
+        return array_combine($timezones, $timezones);
+    }
+
+    public function updateCMSFields(FieldList $fields)
+    {
```
halkyon: brace should be pushed up to the same line as `function`, e.g. `function foo($param) {`

code/SS_Datetimezone.php
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * Adds customisable timezones to the nice method on {@link SS_Datetime}.
+ */
+class SS_Datetimezone extends SS_Datetime
+{
+    /**
+    * Returns the date in the raw SQL-format specific to a given timezone passed from the Member class, e.g. “2006-01-18 16:32:04”
+    */
+    public function Nice() {
+        if($timestamp = $this){
```
halkyon: this is not really good practise, as changing anything that is `$this` (even redefined) will update it by reference, will it not?

Might be a better idea to do something like this:

```$timezone = clone $this;```

Comment (halkyon) [inline code/SS_Datetimezone.php]: You might be better off doing something like this actually:

```
$timestamp = new DateTime($this->value, new DateTimeZone(date_default_timezone_get()));
```

then the whole `if` section is unnecessary. I haven't tested it, but think it should work.

File: code/SS_Datetimezone.php [hunk 6]
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * Adds customisable timezones to the nice method on {@link SS_Datetime}.
+ */
+class SS_Datetimezone extends SS_Datetime {
+	/**
+	* Returns the date in the raw SQL-format specific to a given timezone passed from the Member class, e.g. “2006-01-18 16:32:04”
+	*/
+	public function Nice() {
+		//instantiate new DateTime object based off received timestamp in the default timezone
+		$timestamp = new DateTime($this->value, new DateTimeZone(date_default_timezone_get()));
```
halkyon: This could be tidied up for reduced line lengths and less conditional checking by doing something like the following:

```
public function Nice() {
	$date = new DateTime($this->value);
	$member = Member::currentUser();
	if ($member && $member->exists() && $member->Timezone) {
		$date->setTimezone(new DateTimeZone($member->Timezone));
	}
	return $date->Format('d/m/Y g:ia');
}
```

The `date_default_timezone_get()` isn't necessary either, as the constructor for DateTime will default to the current timezone.

NightJar:
Instead of overriding `SS_Datetime::Nice` it would be better to make this localisation change by overriding `Date::Format` instead, as all functions such as `Nice` use it.

If feeling brave you could also look at `SS_Datetime::FormatFromSettings($member = null)` - but this uses `Zend_Date`.

Also important to keep in mind here with handling input from a user in a different timezone (especially when showing them info in their timezone) is that it is all homogenised to the base timezone. When shown a local date, a local date is likely to be selected when entering e.g. a deployment date/time. Thusly, if this is likely to be a thing (not sure of use-case inspiring this change here) it is also pertinent to look at `setValue`; to convert between input timezone to size's timezone...

...But then that can knock into creating a new input field for dates/times with timezone specified, etc. That or explicitly stating what input should be in. How deep down the rabbit hole does this go? As said, I am unaware of the context, so if this is just on display then no need to read the last two paragraphs (that you just read).

stojg:
Agree, overriding Format would be better.
Btw, is this a good case to use `object::useCustomClass()` to override the class implementation, so we don't have to make custom `LastEdited` and `Created`? Is it a can of worms? Ping @NightJar @halkyon ?

code/extensions/TimezoneMemberExtension.php
```diff
@@ -0,0 +1,27 @@
...
+
+	public function getTimezones() {
+		$timezones = timezone_identifiers_list();
+		return array_combine($timezones, $timezones);
+	}
+
+	public function updateCMSFields(FieldList $fields) {   
+		$fields->removeFieldFromTab('Root', 'Timezone');
+		$field = DropdownField::create(
+		  'Timezone',
```
halkyon: super minor, but the indentation is spaces here

halkyon:
Agreed with @stojg's suggestion to use `Object::useCustomClass()` - this would override all places that use an `SS_Datetime` like for the `Created` and `LastEdited` fields, assuming the core is using `SS_Datetime::create()` everywhere.

code/extensions/TimezoneMemberExtension.php
jakedaleweb: fixed

jakedaleweb:
@halkyon Instead of `Object::useCustomClass()` I've used `Injector` in the YML as per Dylan's suggestion:
https://github.com/silverstripe/deploynaut/pull/664/commits/188c0688aadfad47da1aa1433216f5cbd305558f

PR #667, opened 2016-09-27
Fixing gateway snapshots issues

halkyon:
- Remove -l (user) option for gateway, it's already included in server string
 - Remove -i (identity) option, this file is not on the gateway server


halkyon:
Damn, there are still some problems with this because it passes `-i` to the instance after ssh'ing to the gateway. Closing for now, will re-open later.

PR #669, opened 2016-09-28
First version of the styling for the deployment modal

stojg:


sass/style.sass
```diff
@@ -1264,6 +1267,71 @@ textarea.deploy-key
...
+						&.active
+							border-color: #0685D7
+							color: #505050
+							cursor: pointer
+
+						&.selectable
+							color: #505050
+							cursor: pointer
+
+				.main
```
halkyon: delete?

sass/style.sass
```diff
@@ -1264,6 +1267,71 @@ textarea.deploy-key
...
+				// we remove padding here so the menu can do a left aligned active indicator
+				padding-left: 0
+
+
+				.menu
+					padding-right: 0
+
+					.step
+						font-size: 16px
+						color: #B2B3B4
```
halkyon: are any of these colours going to be re-used? might be worth doing them as variables if so

js/components/Dropdown.jsx
```diff
@@ -34,39 +34,30 @@ var Dropdown = React.createClass({
...
-		this.initSelect2();
-	},
-
 	componentWillUnmount: function() {
 		this.destroySelect2();
 	},
 
 	selector: null,
 
 	initSelect2: function() {
```
halkyon: probably rename this to initSelector or something so it's more generic?

stojg: tru dat

js/components/Dropdown.jsx
```diff
@@ -34,39 +34,30 @@ var Dropdown = React.createClass({
 		this.initSelect2();
 	},
 
-	componentDidUpdate: function() {
-		this.destroySelect2();
-		this.initSelect2();
-	},
-
 	componentWillUnmount: function() {
 		this.destroySelect2();
```
halkyon: Should this be removed too?

stojg: nah, will rename it though. Otherwise the onUpdate event will still hang around even if the unmount the component.

sass/style.sass
stojg: Not sure, I would like to talk to paul a bit before, since I've just picked them from the design and there are colours in the design that isn't referred to in the style guide.

js/components/Dropdown.jsx
stojg: we are actually not using select2 anymore for the modal

sass/style.sass
stojg: deleted.

js/components/Dropdown.jsx
stojg: changed to initSelectize and added code comment

js/components/Dropdown.jsx
stojg: changed to removeSelectize

PR #672, opened 2016-09-28
Log create exceptions

halkyon:


code/jobs/CreateEnvJob.php
```diff
@@ -20,9 +20,12 @@ public function setUp() {
 	 */
 	public function perform() {
 		echo "[-] CreateEnvJob starting" . PHP_EOL;
+
+		$log = new DeploynautLogFile($this->args['logfile']);
```
stojg: lol, I double checked to see if the job actually got logfile set in the creation of the job and it did. we just didn't use it?

code/jobs/CreateEnvJob.php
```diff
@@ -20,9 +20,12 @@ public function setUp() {
...
+		$envCreate = DNCreateEnvironment::get()->byId($this->args['createID']);
+
 		// This is a bit icky, but there is no easy way of capturing a failed deploy by using the PHP Resque
```
stojg: remove the comment, seems like copy pasta?

code/jobs/CreateEnvJob.php
halkyon: yep

code/jobs/CreateEnvJob.php
halkyon: will do

PR #676, opened 2016-09-30
make the deploy modal side bar sticky

stojg:


stojg:
Unfortunately, I couldn't [bootstrap scrollspy](http://getbootstrap.com/javascript/#scrollspy) plugin to work in a modal, will get that working in a separate PR in the future.

PR #681, opened 2016-10-04
Adding approval functionality

halkyon:


js/reducers/approval.js
```diff
@@ -62,20 +61,20 @@ module.exports = function approval(state, action) {
 		case actions.FAIL_APPROVAL_SUBMIT:
 			return state;
 
-		case actions.START_APPROVAL_CANCEL:
+		case actions.SUCCEED_APPROVAL_CANCEL:
 			return _.assign({}, state, {
 				request_sent: false,
 				requested_time: '',
 				approved: false,
```
stojg: I wonder if we should roll `approve`, `bypass` and `rejected` into a singular field together with the deployment.. basically just "expose" the deploy state in the frontend?

halkyon: yeah I think so, we can offload that to `deployment.state` I think, as long as that's the place which is always going to be updated from the server after calling cancel, approve, reject etc.

code/control/ApprovalsDispatcher.php
```diff
@@ -0,0 +1,299 @@
...
+			return $this->getAPIResponse(['message' => 'Method not allowed, requires POST'], 405);
+		}
+
+		$deployment = DNDeployment::get()->byId($request->postVar('id'));
+		$errorResponse = $this->validateDeployment($deployment);
+		if ($errorResponse instanceof \SS_HTTPResponse) {
+			return $errorResponse;
+		}
+
+		$validStates = [DNDeployment::STATE_SUBMITTED, DNDeployment::STATE_NEW];
```
stojg: You should be able to query the statemachine if we can transition, see https://github.com/yohang/Finite/blob/master/src/Finite/StateMachine/StateMachine.php#L146

halkyon: Turns out I didn't need to do this, just check for permissions and then let the transition take care of doing it underneath (it'll throw a bad API response if the state machine errored)

code/model/StateMachineFactory.php
```diff
@@ -13,20 +13,41 @@ public function forDNDeployment(DNDeployment $obj) {
...
 				DNDeployment::STATE_COMPLETED => ['type' => StateInterface::TYPE_FINAL],
 				DNDeployment::STATE_FAILED => ['type' => StateInterface::TYPE_FINAL],
 			],
 			'transitions' => [
+				DNDeployment::TR_NEW => ['from' => [DNDeployment::STATE_SUBMITTED], 'to' => DNDeployment::STATE_NEW],
 				DNDeployment::TR_SUBMIT => ['from' => [DNDeployment::STATE_NEW], 'to' => DNDeployment::STATE_SUBMITTED],
-				DNDeployment::TR_QUEUE => ['from' => [DNDeployment::STATE_SUBMITTED], 'to' => DNDeployment::STATE_QUEUED],
+				DNDeployment::TR_QUEUE => [
+					'from' => [
+						DNDeployment::STATE_SUBMITTED,
```
stojg: what does submitted mean now? Does this mean you can deploy while waiting for approval?

js/containers/buttons/CancelApprovalRequest.jsx
```diff
@@ -5,7 +5,7 @@ var Button = require('../../components/Button.jsx');
 
 const mapStateToProps = function(state) {
 	return {
-		display: state.approval.request_sent && !state.approval.approved,
+		display: !state.deployment.is_loading && state.deployment.data.date_requested && !state.deployment.approved && !state.deployment.rejected,
```
stojg: maybe turn this into a `canCancel(deployState)` func so it's easier on the eyes and the brain

js/containers/buttons/RequestApproval.jsx
```diff
@@ -5,7 +5,7 @@ var Button = require('../../components/Button.jsx');
 
 const mapStateToProps = function(state) {
 	return {
-		display: !state.approval.request_sent && state.approval.approved_by && state.git.selected_ref,
+		display: !state.deployment.is_loading && !state.deployment.data.date_requested && state.deployment.approver_id && state.git.selected_ref,
```
stojg: maybe turn this into a func so it's easier on the eyes and brain (just for the state.deployment parts)

code/model/StateMachineFactory.php
halkyon: submitted is awaiting approval

halkyon: removed submitted => queued, that is not a transition we support

PR #687, opened 2016-10-06
Add a loading indicator compononent

stojg:


js/components/LoadingBar.jsx
```diff
@@ -0,0 +1,23 @@
...
+
+const LoadingBar = function(props) {
+
+	if (!props.show) {
+		return null
+	}
+
+	return (
+		<div className="loading-horisontal">
+			<div id="loading-horisontal-1"></div>
```
halkyon: should be `horizontal`

stojg: hahaha.. screw you English language!

halkyon: annoyingly inconsistent language as usual: `apologized` and `apologised` are valid, but `horisontal` is not!

PR #689, opened 2016-10-06
Fix navigation

stojg:


js/containers/DeployHistory.jsx
```diff
@@ -27,8 +27,9 @@ const DeployHistory = function(props) {
 			<div className="text-center">
 				<nav>
 					<Pagination
-						prev
```
halkyon: Why first and last only?

stojg:
Mimics current look

halkyon:
@stojg Mimics, sure, but that's going to mean it goes from 1 straight to 6 if you click the right arrow... wouldn't it make more sense if you left them with `next` and `prev`?

stojg:
Sorry, I don't understand
Current:
image
My suggestion:
image

Unfortunately, there doesn't seem to be an easy way to get the last page button from the lib.

If you click for example page 5 you get

image
So in my humble opinion, we don't need the prev and next

PR #697, opened 2016-10-10
Log state transitions in deploy log

halkyon:


code/model/StateMachineFactory.php
```diff
@@ -77,6 +77,12 @@ protected function addHandlers($machine) {
 		$class = get_class($machine->getObject());
 		if (empty($this->config()->handlers[$class])) return;
 
+		// this is the "internal" state listener that logs all state transitions into a log file for debugging
+		$machine->getDispatcher()->addListener(
+			'finite.post_transition',
+			[Injector::inst()->get('DNDeploymentHandlers'), 'onAnyTransition']
```
stojg: maybe call it `onAfterTransition` if we follow SS jargon and make clear to users just looking at that method that the transition has already been done?

halkyon: Fixed!

PR #715, opened 2016-10-14
add getBaseDir and check/set before deploy via capistrano

andrewandante:
Currently causing capistrano deployments to fail when using PackageCache because basedir isn't being set.


halkyon:
Something's screwed up with the config YAML, as it was supposed to set this: 

https://github.com/silverstripe/deploynaut/blob/master/_config/dnroot.yml#L33

halkyon:
Isn't needed as it's a misconfiguration of the project code. We can't hardcode that path in `SizeRestrictedPackageCache` anyway. I'll close this for now.

PR #751, opened 2016-11-04
changes to make vstacks appear under their base in /naut/projects

jakedaleweb:


templates/Layout/DNRoot_projects.ss
```diff
@@ -29,47 +28,78 @@
...
+											<img src="$Icon" />
+											<a href="$URL">View code<i class="fa fa-external-link-square superscript"></i></a>
+										</div>
+									<% end_with %>
+								<% else %>
+									-
+								<% end_if %>
+							</td>
+						</tr>
+						<% if VirtualSelectMenu($ID) %>
```
halkyon: Hmm... this `VirtualSelectMenu` doesn't exist in the deploynaut module does it?

jakedaleweb: Nope, that's over here:
https://github.com/silverstripeltd/platform-dashboard/pull/330/files

Is that going to cause issues?

halkyon: It's currently in an extension in the project code, introducing it here in the module may not be the best idea. Maybe you should override this template into the project theme instead?

PR #757, opened 2016-11-07
Remove unncessary fetch of upcoming deployments

stojg:
The data can be filtered out of the deployment history

The front end is already setup to pick out the relevant deployments from the list of all deployments. This change will get rid of an unnecessary call to the backend.


code/model/DNEnvironment.php
```diff
@@ -680,33 +680,19 @@ public function DeployHistory($orderBy = '') {
...
 		$sort['LastEdited'] = 'DESC';
 
-		return $this->Deployments()
+
+		$deployments = $this->Deployments()
 			->where('"SHA" IS NOT NULL')
-			->filter('State', [
+			->sort($sort);
+
+		if($this->IsNewDeployEnabled()){
```
halkyon: `if ($this->IsNewDeployEnabled()) {` (minor spacing ;-))

PR #766, opened 2016-11-07
background updates of deployment list

stojg:


code/control/DeployDispatcher.php
```diff
@@ -77,6 +77,13 @@ public function history(\SS_HTTPRequest $request) {
 
 		$list = $this->environment->DeployHistory('DeployStarted');
 
+		$fromTimestamp = $request->requestVar('from');
+		if($fromTimestamp) {
```
halkyon: you know what i will say here about code formatting? ;-)

PR #806, opened 2016-11-10
Perfomance tweaks

stojg:
The ApiFormatter is run in a tight foreach loop, often from the DeployDispatcher,
this changes adds per request cache for often accessed data, like Member.
Locally (without any OPcache) is see a performance improvement of 50%

### DIFF
--- a/code/api/DeploynautAPIFormatter.php
+++ b/code/api/DeploynautAPIFormatter.php
@@ -9,6 +9,12 @@ class DeploynautAPIFormatter {
 	private static $_cache_project_members = null;
 
 	/**
+	 * This is a per request cache of Members
+	 * @var null|array
+	 */
+	private static $_cache_members = null;
+
+	/**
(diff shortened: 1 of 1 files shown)


code/api/DeploynautAPIFormatter.php
```diff
@@ -117,31 +117,43 @@ public function getDeploymentData(\DNDeployment $deployment) {
...
 	 * user who can create the deployment but is not part of the stack roles.
 	 *
 	 * @param \DNProject $project
-	 * @param \Member $member
-	 * @return array
+	 * @param int $memberID
+	 * @return null|array
 	 */
-	public function getStackMemberData(\DNProject $project, \Member $member) {
+	public function getStackMemberData(\DNProject $project, $memberID) {
```
stojg: by passing the MemberID here we don't need to initialise a full Member Object if we cached it further down

PR #811, opened 2016-11-10
Copy selected options when reducing state for deploy options

halkyon:


js/reducers/git.js
```diff
@@ -41,7 +41,10 @@ module.exports = function git(state, action) {
 				selected_name: ""
 			});
 		case actions.TOGGLE_OPTION: {
-			const selected_options = state.selected_options;
```
stojg: does `const selected_options = _.assign({}, state.selected_options);` work as short hand?

PR #813, opened 2016-11-11
style and set status update during/after deployment

stojg:


sass/components/_deployment-status-bar.sass
```diff
@@ -0,0 +1,46 @@
...
+	.deployment-status-icon
+		font-size: 50px
+		line-height: 80px
+		padding: 0 20px
+		float: left
+		color: $platform-blue
+
+		&.Aborting
+			color: $status-aborting
+		&.Pending
```
halkyon: there's no Pending status :)

PR #848, opened 2016-11-24
dont create or update deployment on approval submit

stojg:


stojg:
The deploy should already have been created when continuing from the deployment plan step

PR #860, opened 2016-11-28
add a deleted state for deployments for ui updates and auditing

stojg:
By using a soft delete approach we still keep the data for auditing
and can send 'real time' updates to the UI about which deployment(s)
was deleted for other users having the same page opened.

### DIFF
--- a/code/control/DeployDispatcher.php
+++ b/code/control/DeployDispatcher.php
@@ -122,11 +122,16 @@ class DeployDispatcher extends Dispatcher {
 			return $errorResponse;
 		}
 
-		$id = $deployment->ID;
-		$deployment->delete();
+		try {
+			$deployment->getMachine()->apply(\DNDeployment::TR_DELETE);
+		} catch (\Exception $e) {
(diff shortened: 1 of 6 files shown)


halkyon:
Won't this show in history still? you may have to filter it out in the `DeployHistory()` method on server-side?

stojg:
The history only shows 
```
				case deployStates.STATE_COMPLETED:
				case deployStates.STATE_INVALID:
				case deployStates.STATE_FAILED:
```

halkyon:
Yeah, it's more the perspective of the API never returning a deleted state or something.

stojg:
I'm not sure how to solve 'the API never returning a deleted state' without setting up new actions, new XHR requests, reducers etc.

stojg:
The "old UI" checks out okay since it only returns if the flag is off
```
DNDeployment::STATE_COMPLETED,
DNDeployment::STATE_FAILED,
DNDeployment::STATE_INVALID
```

halkyon:
ah okay, all good then

PR #867, opened 2016-11-29
randomise loading messages

stojg:


stojg:
supersede https://github.com/silverstripe/deploynaut/pull/857/

js/containers/buttons/SaveTargetRelease.jsx
```diff
@@ -2,9 +2,13 @@ var ReactRedux = require('react-redux');
...
 var actions = require('../../_actions.js');
 var Button = require('../../components/Button.jsx');
+var message = require('../../constants/messages.js');
+
+// keep it the same between render calls by deciding the random message on import
+const loadingMessage = message.getRandom();
 
 const mapStateToProps = function(state) {
-	const text = state.plan.is_loading ? 'Reticulating splines' : 'Continue';
+	const text = state.plan.is_loading ?  loadingMessage : 'Continue';
```
halkyon: SUPER CRITICAL ISSUE. ` loadingMessage` instead of `loadingMessage`

stojg: wut?

halkyon:
could you also do this for "Requesting abort" on the abort button? :)

js/containers/buttons/SaveTargetRelease.jsx
stojg: ah, ws

stojg:
I tried adding it to the abort button, but it makes less sense since it's a "momentary" button that then goes back to the 'Abort' state. I also don't want to joke around with abort, since the user is probably a bit stressed at that point.

PR #876, opened 2017-01-08
lock gridfield extensions to 1.4.2 instead of using dev-master

stojg:


stojg:
dev-master of gridfield-extensions are now requiring silverstripe framework 4.0

stojg:
all good, we should have locked it down a while ago :P

PR #877, opened 2017-01-12
MINOR helper fuctions for CMS sanity

andrewandante:
Tidies up a few CMS things when combined with the Dashboard PR:

https://github.com/silverstripeltd/platform-dashboard/pull/362

### DIFF
--- a/code/model/DNProject.php
+++ b/code/model/DNProject.php
@@ -213,6 +213,15 @@ class DNProject extends DataObject {
 	}
 
 	/**
+	 * @return string
+	 */
+	public function getPrimaryEnvType() {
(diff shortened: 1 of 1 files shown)


stojg:
I recommend that you use a CMS extension to modify the DNAdmin ModelAdmin, since this module should not have a coupling to other modules. 

See https://docs.silverstripe.org/en/3/developer_guides/customising_the_admin_interface/how_tos/extending_an_existing_modeladmin/#extending-existing-modeladmin

andrewandante:
Agreed - have moved most of the code out into an extension. Think the remaining two functions make sense on `DNProject` though.

